PR #1, opened 2019-12-27
PlaceDetails UI (first draft)

mredigonda:

This is a work in progress with the goal of solving everything UI-related for the PlaceDetails component. In this first part of developing the PlaceDetails component, I'll be using mocked data, and later on I'll discover how to fetch real user data with an orchestrator.

The purpose of sharing the code early is to receive feedback given that I'm just starting out contributing code.

[Workspace](https://maxi--checkoutio.myvtex.com/cart)

The code that this PlaceDetails component tries to replace is [in the address-form repository](https://github.com/vtex/address-form/blob/master/react/AddressSummary.js).

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [ ] Updated `CHANGELOG.md`.

Currently, the component looks like this:
Screen Shot 2019-12-27 at 7 36 12 PM

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements 

#### Notes

The original code uses two `reduce` calls to (I believe) remove nulls, and then to add `<br />` tags, but I'm ignoring nulls since they don't get rendered, and it seems that they were only being removed to add the `<br />` tags, which I'm doing in a different way.

Things to solve:

- [x] Use proper names (LineComponent?, CountryDescription?)
- [ ] Discover if maskedInfoIcon and country should be rendered as in the original component
- [x] Make sure all classes are applied as in the original component


react/PlaceDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+    summary: [
+      [{ name: 'street' }, { delimiter: ', ', name: 'complement' }],
+      [
+        { name: 'city' },
+        { delimiter: ', ', name: 'state' },
+        { delimiter: ' ', name: 'postalCode' },
+      ],
+    ],
+  },
+]
```
marcoskwkm: I would move all this mock data to a separate file.

react/PlaceDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+          { value: 'KOR', label: 'Korea' },
+        ]}
+        value={option}
+        onChange={(_: any, newVal: string) => {
+          setOption(newVal)
+          let newSummary = countryDescriptions.find(
+            description => description.name == newVal
+          )
+          setSummary(newSummary ? newSummary.summary : [])
+        }}
```
marcoskwkm: I would move this to separate function to improve code readability.

react/PlaceDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+                  <span className={field.name + '-delimiter-after'}>
+                    {field.delimiterAfter}
+                  </span>
+                )}
+              </span>
+            ) : null
+          }
+        ),
+        <br className={'line' + (index + 1) + '-delimiter'} key={index} />,
+      ])}
```
marcoskwkm: This is really hard to read and understand what's going on. Break it into smaller variables/functions.

react/PlaceDetails.tsx
```diff
@@ -0,0 +1,134 @@
...
+        { delimiter: ', ', name: 'state' },
+        { delimiter: ' ', name: 'postalCode' },
+      ],
+    ],
+  },
+]
+
+const PlaceDetails: StorefrontFunctionComponent<PlaceDetailsProps> = () => {
+  const [option, setOption] = useState<string>('ARG')
+  const [summary, setSummary] = useState<LineComponent[][]>([])
```
marcoskwkm: Note that you're setting `ARG` as initial value of `option` but setting the initial value of `summary` as empty. This way when I load the page I see `Argentina` in the dropdown but I only see the address once I select a different country in the dropdown and then select Argentina again.

react/PlaceDetails.tsx
mredigonda: Done.
Is the new file name (and location) correct? Do you guys have any convention on this?

mredigonda: In the future that file will almost certainly disappear (and I'd use real data instead), but it will be useful to know for future occasions.

react/PlaceDetails.tsx
mredigonda: Great, I decided to stop using a default value so that I don't have to bother getting the summary for that default country.

react/PlaceDetails.tsx
marcoskwkm: You can put the mock data in a file in the `react/__mocks__` folder, so in the future you can use the same data to write unit tests ; )

react/PlaceDetails.tsx
mredigonda: I moved it to a function outside, but still inside the PlaceDetails function component to be able to use the setOption and setSummary functions.

react/PlaceDetails.tsx
mredigonda: Awesome, I'll do that!

react/PlaceDetails.tsx
mredigonda: I copied this part from the previous version and I didn't want to my significant changes in the logic until I know it won't break anything, having said that, I broke it down a bit, although it's still quite cumbersome. I plan to get back to this once I understand exactly how it works so that I can make a simpler version of this. Thanks!

react/PlaceDetails.tsx
```diff
@@ -0,0 +1,87 @@
...
+  type: 'object',
+  properties: {
+    title: {
+      title: 'editor.base-store-component.title.title',
+      description: 'editor.base-store-component.title.description',
+      type: 'string',
+      default: null,
+    },
+  },
+}
```
marcoskwkm: We can remove this for now.

jeffersontuc:
Just a tip: You can add to your workspace url a query string to add items to cart.

e.g .  https://maxi--checkoutio.myvtex.com/cart/add?sku=288

So it makes easier to anyone who is reviewing your PR.

PR #5, opened 2020-01-27
AddressForm, new JSON format, internationalization, and validations

mredigonda:

I was using an outdated JSON file to get country-specific rules, this PR is to update that data to a new format that will be more convenient for the future.

#### Added

- `LocaleSwitcher` dependency to handle locale query strings
- Internationalization
- Max length, and required validations
- Added mock for `AddressContext`
- Added JSON in new format for `ARG`, `BRA` and `USA`, and its TS interface definitions

#### Changed

- `AddressForm` now contains an instance of `PlaceDetails`, and an edit button
- `PlaceDetails` and `AddressForm` now consume `address` information directly from `AddressContext` orchestrator

[Workspace](https://maxi--checkoutio.myvtex.com/places)

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
_ | New feature 
_ | Breaking change 
_ | Technical improvements 

#### Notes

There is some work to be done in the tests yet.


react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+
+  const getSummaryFields = (summary: LineFragment[][]) => {
+    let summaryFields = new Set()
+    summary.forEach((line: LineFragment[]) => {
+      line.forEach((fragment: LineFragment) => {
+        summaryFields.add(fragment.name)
+      })
+    })
+    return summaryFields
+  }
```
jeffersontuc: What do you think about moving this function outside the component declaration so, that way, the function isn't defined every time the component is rendered or updated?

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
-const AddressForm: StorefrontFunctionComponent<Props> = ({
-  address,
-  summary,
-}) => {
-  const { setAddress } = useAddressContext()
+  const hasOptions = (field: OptionsField) => {
+    return field && field.options
+  }
 
   const parseLineFragment = (fragment: LineFragment) => {
```
jeffersontuc: Remember that in a Functional Component the `render` method is its entire body. So every time the component is rendered, the functions are defined again which can compromise your component performance.

CHANGELOG.md
```diff
@@ -7,9 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Added mock for `AddressContext`
+- Added JSON in new format for `ARG`, `BRA` and `USA`, and their TS interface definitions
+
+### Changed
+
+- `AddressForm` now contains an instance of `PlaceDetails`, and an edit button
+- `PlaceDetails` and `AddressForm` now consume `address` information directly from `AddressContext` orchestrator
+- Simplified `PlaceDetails` tests
+
+## [0.0.4] - 2020-01-17
```
lucasecdb: did you accidentally change this date? the diff above says the `v0.0.4` was released on `2020-01-23` instead of `2020-01-17`, which is the date `v0.0.3` was released

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+  },
+  ['number-option']: {
+    defaultMessage: '',
+    id: 'place-components.label.number-option',
+  },
+  complement: {
+    defaultMessage: '',
+    id: 'place-components.label.complement',
+  },
+  stateAbbreviation: {
```
darlenedms: Why some keys are `kebab-case` and others `camelCase`?

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+import rules from './countries/rules'
+import PlaceDetails from './PlaceDetails'
+import { FormattedMessage, defineMessages } from 'react-intl'
+import NumberOption from './components/NumberOption'
 
-interface Props {
-  address: Address
-  summary: LineFragment[][]
-}
+defineMessages({
```
lucasecdb: shouldn't you use the return value of this function?

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+  },
+  stateAbbreviation: {
+    defaultMessage: '',
+    id: 'place-components.label.stateAbbreviation',
+  },
+})
+
+const AddressForm: StorefrontFunctionComponent<{}> = () => {
+  const { address, setAddress } = useAddressContext()
+  const [editing, setEditing] = useState<boolean>(false)
```
lucasecdb: nit: TS can infer that the generics here is a boolean, so you can leave out the parameter

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+  }
+
+  const displayMode: keyof Display = editing ? 'minimal' : 'compact'
+
   return (
     <div>
+      <PlaceDetails display={displayMode} />
+      {!editing && (
+        <ButtonPlain onClick={onEditButtonClick} title="edit">
+          <FormattedMessage id="place-components.label.edit" />
```
lucasecdb: you should use the return value from the `defineMessages` here

react/__tests__/AddressForm.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+      queryByLabelText,
+      queryByDisplayValue,
+      debug,
+    } = renderComponent(address)
+    fireEvent.click(queryByText('Edit'))
+    fireEvent.click(queryByLabelText('Without number'))
+    console.log(queryByLabelText('Without number'))
+    console.log(queryByLabelText('Without number').click())
+
+    debug()
```
lucasecdb: 🤔

react/typings/countryRulesTypes.d.tsx
```diff
@@ -0,0 +1,48 @@
+import { Address } from 'vtex.checkout-graphql'
```
marcoskwkm: Does this file need to have `.d.tsx` extension instead of just `d.ts`?

react/components/NumberOption.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return (
+    <span>
+      <span className="w-25 dib mh3">
+        <Input {...getInputProps()} />
+      </span>
+      {showCheckbox && (
+        <span className="w-25 dib mh3">
+          <Checkbox {...getCheckboxProps()} />
```
lucasecdb: can't you just pass the props directly here? to avoid having unecessary functions declared in the body of the component

react/components/NumberOption.tsx
```diff
@@ -0,0 +1,99 @@
...
+      label: <FormattedMessage id="place-components.label.without-number" />,
+      onChange,
+      checked: disabled,
+    }
+  }
+
+  return (
+    <span>
+      <span className="w-25 dib mh3">
+        <Input {...getInputProps()} />
```
lucasecdb: same here

react/countries/USA.ts
```diff
@@ -0,0 +1,141 @@
...
+    street: {
+      label: 'addressLine1',
+      required: true,
+      size: 'xlarge',
+    },
+    number: {
+      label: 'number',
+      maxLength: 750,
+      hidden: true,
+      // defaultValue: 'N/A',
```
marcoskwkm: Are you intentionally leaving this commented line?

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+  }
 
   const parseLineFragment = (fragment: LineFragment) => {
-    return address[fragment.name] != null ? (
+    const field = fields[fragment.name as keyof Fields]
+    const labelName = field && field.label ? field.label : null
+
+    if (ignoredFields.has(fragment.name)) return null
+    if (address[fragment.name] == null) return null
+    if (numberHasWithoutOption(labelName)) return <NumberOption showCheckbox />
```
darlenedms: Wouldn't be clearer to name this function `hasWithoutNumberOption`?
```suggestion
    if (hasWithoutNumberOption(labelName)) return <NumberOption showCheckbox />
```

react/countries/BRA.ts
```diff
@@ -0,0 +1,121 @@
...
+      optionsCaption: 'UF',
+      options: [
+        { label: 'Acre', value: 'AC' },
+        { label: 'Alagoas', value: 'AL' },
+        { label: 'Amapá', value: 'AP' },
+        { label: 'Amazonas', value: 'AM' },
+        { label: 'Bahia', value: 'BA' },
+        { label: 'Ceará', value: 'CE' },
+        { label: 'Distrito Federal', value: 'DF' },
+        { label: 'Espírito Santo	', value: 'ES' },
```
marcoskwkm: There's some extra space after `Santo`.

react/countries/BRA.ts
```diff
@@ -0,0 +1,121 @@
...
+        { label: 'Piauí', value: 'PI' },
+        { label: 'Rio de Janeiro', value: 'RJ' },
+        { label: 'Rio Grande do Norte', value: 'RN' },
+        { label: 'Rio Grande do Sul', value: 'RS' },
+        { label: 'Rondônia', value: 'RO' },
+        { label: 'Roraima', value: 'RR' },
+        { label: 'Santa Catarina', value: 'SC' },
+        { label: 'São Paulo', value: 'SP' },
+        { label: 'Sergipe', value: 'SE' },
+        { label: 'Tocantins	', value: 'TO' },
```
marcoskwkm: ditto

react/AddressForm.tsx
```diff
@@ -1,46 +1,162 @@
...
+
+    const getFieldProps = (fragment: LineFragment) => {
+      const maxLength = field && field.maxLength ? field.maxLength : null
+      const autoComplete =
+        field && field.autoComplete ? field.autoComplete : null
+      const required = field && field.required ? field.required : null
+      const label = (
+        <FormattedMessage id={`place-components.label.${labelName}`} />
```
darlenedms: An chance of `labelName` to be `null`? (according to https://github.com/vtex-apps/place-components/pull/5/files#diff-50aad4c7907d795061b7b081a887f18bR82)

react/components/NumberOption.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    return {
+      label,
+      value,
+      onChange,
+      disabled,
+      ...(maxLength && { maxLength }),
+      ...(autoComplete && { autoComplete }),
+      ...(required && address['number'].length == 0 && fieldRequired),
```
marcoskwkm: Use `===` here. It is safer since `foo == 0` evaluates to `true` also when `foo` is `false` or `''` .

react/components/NumberOption.tsx
```diff
@@ -0,0 +1,99 @@
...
+}) => {
+  const { address, setAddress } = useAddressContext()
+  const [disabled, setDisabled] = useState(false)
+
+  const getInputProps = () => {
+    const field = rules[address.country].fields['number']
+    const maxLength = field && field.maxLength ? field.maxLength : null
+    const autoComplete = field && field.autoComplete ? field.autoComplete : null
+    const required = field && field.required ? field.required : null
+    const labelName = field && field.label ? field.label : null
```
marcoskwkm: I think it is more idiomatic to write this and the next lines as
`const maxLength = field?.maxLength ?? null`

react/__tests__/AddressForm.test.tsx
mredigonda: Yap, tests are a work in progress yet :sweat_smile:

react/components/NumberOption.tsx
mredigonda: Somehow my VSCode is not understanding that syntax of `?.` and it's breaking all other syntax checks, that's why I avoided it, I'll try again though.

marcoskwkm: You might need to update the versions of the linting packages in your `package.json`. Try updating     `eslint`, `eslint-config-vtex-react` and `prettier` and see if VSCode still complaints.

CHANGELOG.md
mredigonda: It was on purpose, that `v0.0.4` previously released was actually empty.

react/AddressForm.tsx
mredigonda: Great, removed that!

react/AddressForm.tsx
mredigonda: LabelName can't really be null, I just messed with typescript, and we already fixed this with @lucasecdb

react/AddressForm.tsx
mredigonda: Seems like a good idea since that function doesn't use any state variables :+1:

react/AddressForm.tsx
mredigonda: Yes, it would 😅

react/countries/USA.ts
mredigonda: It was unintentional, this line should be deleted, it was inherited from the previous component, but according to Fernandinho, it's safe to remove it.

react/AddressForm.tsx
mredigonda: We reviewed this together, please confirm it's okay now :+1:

react/AddressForm.tsx
mredigonda: I'm using `defineMessages` to handle dynamic messages, in this case, I think the current way suffices. Would it be more desirable to add this to the `defineMessages` object?

lucasecdb: sure, if you are already using `defineMessages` in this file I think it would be better if all messages for this file were defined there.

react/AddressForm.tsx
```diff
@@ -1,46 +1,158 @@
...
+    getSummaryFields(display.compact)
+  )
+
+  const hasWithoutNumberOption = (label: string) => {
+    return label.length >= 6 && label.substr(label.length - 6) == 'Option'
+  }
+
+  const hasOptions = (field: OptionsField) => {
+    return field.options
+  }
```
lucasecdb: can you move this functions to outside the component body?

react/AddressForm.tsx
```diff
@@ -1,46 +1,158 @@
...
+        ) {
+          setAddress({
+            ...address,
+            [fragment.name]: event.target.value,
+          })
+        }
+      }
+      const fieldRequired = {
+        errorMessage: (
+          <FormattedMessage id={`place-components.error.fieldRequired`} />
```
lucasecdb: nit
```suggestion
          <FormattedMessage id="place-components.error.fieldRequired" />
```

react/AddressForm.tsx
```diff
@@ -1,46 +1,158 @@
...
+    if (hasWithoutNumberOption(labelName)) return <NumberOption showCheckbox />
+
+    const getFieldProps = (fragment: LineFragment) => {
+      const { maxLength, autoComplete, required } = field
+      const label = <FormattedMessage {...messages[labelName]} />
+
+      const onChange = (event: React.ChangeEvent) => {
+        if (
+          event.target instanceof HTMLInputElement ||
+          event.target instanceof HTMLSelectElement
```
lucasecdb: why do you need to do this check? is there a chance this `onChange` will be called from other elements?

react/AddressForm.tsx
```diff
@@ -1,46 +1,158 @@
...
+  }
 
   const parseLineFragment = (fragment: LineFragment) => {
-    return address[fragment.name] != null ? (
+    const field = fields[fragment.name as keyof Fields]
+    if (!field) return null
+    const labelName: LabelType = field.label as LabelType
+
+    if (ignoredFields.has(fragment.name)) return null
+    if (address[fragment.name] == null) return null
```
lucasecdb: can you group these `if`s to a single one? such as 

```js
if (!field || ignoredFields.has(fragment.name) || address[fragment.name] == null) {
  return null
}
```

react/AddressForm.tsx
```diff
@@ -1,46 +1,156 @@
...
+      })
+    }
+    const fieldRequired = {
+      errorMessage: <FormattedMessage {...messages.fieldRequired} />,
+    }
+    const value = address[fragment.name]
+
+    return {
+      label: (
+        <FormattedMessage {...messages[label as keyof (typeof messages)]} />
```
mredigonda: @lucasecdb using `typeof messages` in the console only returns `object`. Would it be better to directly use `keyof object` in that case? or is there a way to define the type of that label more specifically?

lucasecdb: this is for typescript only, so it will correctly infer the type from the `messages` variable. in javascript land it will still be a regular object, but typescript will be able to tell the specific type

think doing `typeof SomeReactComponent`. in javascript, it will output `object`, but TS will be able to see it as a React component, that could include methods such as `componentDidMount`, etc

lucasecdb: you can create a [type alias](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) so it will store the actual type of the variable.

```ts
type LabelType = keyof (typeof messages)
```

mredigonda: Oh, great then, I removed the type alias since it was only used once, but if you think it's cleaner I can add it again, thank you!

react/AddressForm.tsx
mredigonda: Great, I'll have this in mind for future components, and will see how to improve the existing components. Currently, there are some nesting of functions that use some variables from the parent function, but after talking to Marcos, I think that those should either be small, or be outside. Thanks!

react/AddressForm.tsx
lucasecdb: ok! no problem

react/AddressForm.tsx
```diff
@@ -1,49 +1,165 @@
...
+          <FormattedMessage {...messages.edit} />
+        </ButtonPlain>
+      )}
       <div>{summary.map(parseLine)}</div>
     </div>
   )
 }
 
-export default AddressForm
+export default injectIntl(AddressForm)
```
lucasecdb: I think you can use the [`useIntl`](https://github.com/formatjs/react-intl/blob/master/docs/API.md#useintl-hook) hook from `react-intl`, if you prefer that

react/__tests__/AddressForm.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    expect(getByText(address.state as string)).toBeTruthy()
+    expect(getByText(address.postalCode as string)).toBeTruthy()
+
+    expect(queryByDisplayValue(address.street as string)).toBeFalsy()
+    expect(queryByDisplayValue(address.number as string)).toBeFalsy()
+    expect(queryByDisplayValue(address.neighborhood as string)).toBeFalsy()
+    expect(queryByDisplayValue(address.city as string)).toBeFalsy()
+    expect(queryByDisplayValue(address.state as string)).toBeFalsy()
+    expect(queryByDisplayValue(address.postalCode as string)).toBeFalsy()
+  })
```
lucasecdb: the test name says it checks the inputs to be `read-only`, but there's nothing asserting that in the function. are you still working on these tests?

react/__tests__/AddressForm.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should have an edit button', () => {
+    const { getByText } = renderComponent(address)
+    expect(getByText('Edit')).toBeTruthy()
```
lucasecdb: I think this is more idiomatic
```suggestion
    expect(getByText('Edit')).toBeInTheDocument()
```

react/typings/countryRulesTypes.d.ts
```diff
@@ -5,3 +5,38 @@ export interface LineFragment {
...
+
+export interface Field {
+  label: string
+  hidden?: boolean
+  maxLength?: number
+  size?: string
+  required?: boolean
+  autoComplete?: string
+  optionsCaption?: string
+  options?: ({ label: string; value: string })[]
```
lucasecdb: you can also declare this as `Array<{ label: string; value: string }>` if you prefer that

react/__tests__/AddressForm.test.tsx
lucasecdb: you could also change the other assertions that were using `toBeTruthy` and `toBeFalsy` to be `toBeInTheDocument` and `not.toBeInTheDocument`, which would be a lot better for readability

react/typings/countryRulesTypes.d.ts
mredigonda: It seems that this is prohibited by our linter: `Array type using 'Array<T>' is forbidden. Use 'T[]' instead` 😢

react/typings/countryRulesTypes.d.ts
```diff
@@ -5,3 +5,38 @@ export interface LineFragment {
...
+
+export interface Field {
+  label: string
+  hidden?: boolean
+  maxLength?: number
+  size?: string
+  required?: boolean
+  autoComplete?: string
+  optionsCaption?: string
+  options?: { label: string; value: string }[]
```
jeffersontuc: Why not

```ts
export interface Field {
    ...
    options?: Option[]
    ...
}

export interface Option {
    label: string
    value: string
}
```
?

PR #6, opened 2020-02-12
Location select location country

mredigonda:

This PR adds the `LocationCountry` and the `LocationSelect` components to the PlaceComponents repository. It also adds the JSON country object for Bolivia (to test `LocationSelect`), and it adds max/min widths to form inputs of the `AddressForm` component.

[Workspace](https://loccountrylocselect--checkoutio.myvtex.com/places)

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- LocationCountry
+  - Guesses country from store settings
```
darlenedms:
```suggestion
- Guesses country from store settings
```
Is this intentional?

react/AddressForm.tsx
```diff
@@ -108,30 +108,42 @@ const AddressForm: StorefrontFunctionComponent = () => {
...
 
+    const style = styleRules[field.label]
+
     if (hasWithoutNumberOption(field.label))
-      return <NumberOption showCheckbox />
+      return (
+        <div className="flex-auto mh3" style={style as React.CSSProperties}>
+          <NumberOption showCheckbox />
+        </div>
+      )
```
darlenedms: Missing curly braces is kinda weird with multiple lines 🤔

marcoskwkm: Always use curly braces in `if`  statements, even if the body has a single statement, as that avoids bugs in case someone adds something to this `if` block in the future.

react/AddressForm.tsx
```diff
@@ -108,30 +108,42 @@ const AddressForm: StorefrontFunctionComponent = () => {
 
   const parseLineFragment = (fragment: LineFragment) => {
     const field = fields[fragment.name as keyof Fields]
+
     if (
       !field ||
       ignoredFields.has(fragment.name) ||
       address[fragment.name] == null
     )
       return null
```
marcoskwkm: Add curly braces.

react/LocationCountry.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const LocationCountry: StorefrontFunctionComponent<{}> = () => {
+  const intl = useIntl()
+  const { address, setAddress, countries } = useAddressContext()
+  const {
+    culture: { country: storeCountry },
+  } = useRuntime()
+  let { country } = address
+
+  if (country && countries && !countries.includes(country))
```
marcoskwkm: Curly braces.

react/LocationCountry.tsx
```diff
@@ -0,0 +1,91 @@
...
+        number: null,
+        postalCode: null,
+        state: null,
+        street: null,
+      })
+    },
+    options,
+  }
+
+  return <Dropdown {...dropdownProps}></Dropdown>
```
marcoskwkm: I think you can do `<Dropdown {...dropdownProps} />`

react/LocationCountry.tsx
```diff
@@ -0,0 +1,91 @@
...
+const LocationCountry: StorefrontFunctionComponent<{}> = () => {
+  const intl = useIntl()
+  const { address, setAddress, countries } = useAddressContext()
+  const {
+    culture: { country: storeCountry },
+  } = useRuntime()
+  let { country } = address
+
+  if (country && countries && !countries.includes(country))
+    throw "The country from that address doesn't belong to the country list, can't render LocationCountry"
```
marcoskwkm: How about making it explicit which `country` is not in the list?

react/LocationSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+    id: 'place-components.label.locality',
+  },
+})
+
+const LocationSelect: StorefrontFunctionComponent<{}> = () => {
+  const { address, setAddress } = useAddressContext()
+  const countryRules = rules[address.country]
+
+  if (!countryRules.locationSelect)
+    throw 'The LocationSelect component is not applicable to this country :('
```
marcoskwkm: Curly braces.

react/LocationSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const { countryData, fields } = countryRules.locationSelect
+
+  const getCompletedFieldsArray = () => {
+    let completedFields = []
```
marcoskwkm: This can be a `const`.

react/LocationSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+    for (let field of fields) {
+      if (address[field.name]) {
+        completedFields.push(address[field.name])
+      } else {
+        break
+      }
+    }
+    return completedFields
+  }
```
marcoskwkm: Can't you rewrite this as a `map-filter`? I tend to prefer functional style because it describes more clearly what you're doing.

CHANGELOG.md
mredigonda: Yes, this was intentional because it is `LocationCountry` who does that, but I thought I was going to add other points, I could rewrite it better 👍

react/AddressForm.tsx
mredigonda: I used to always add curly braces to `if` statements for that exact reason 😅 I'll fix it.

react/LocationSelect.tsx
mredigonda: I read that it's not a good idea to use `const` for non-primitive types (objects, arrays, functions), since one would expect that a `const` variable doesn't change its value, but the `const` keyword only adds the restriction that the variable can't be reassigned (so if it is an object/array/function, its value can still change).
What do you think about this? It seems that the pattern we use is using `const` when it's a variable we only use once.

react/LocationSelect.tsx
mredigonda: I don't know how to do this with `map-filter` directly, since the filtering would be kind of strange (I need to stop taking elements after I find the first `null` element).
I believe the functional way to do this would be with a `takeWhile` function, this is the best I could get;
```typescript
const takeWhile = (array: any[], f: (arg0: any) => boolean) => {
  const first = array.findIndex(f)
  return first === -1 ? array : array.slice(0, first)
}
const addressFields = fields.map(field => address[field.name])
return takeWhile(addressFields, x => !x)
```
What do you think about this?

react/LocationSelect.tsx
marcoskwkm: The argument that `const` could be misleading in that the variable can change makes sense, and I'd like to read about it if you still have the source, but I still think a `const` is better here. The reason is exactly what you said, it prevents the variable from being reassigned. Variable reassigning is problematic because sometimes it means you're using the same variable for different things, which can lead to bugs and a code that is harder to understand.

react/LocationSelect.tsx
marcoskwkm: Sorry, I completely missed the `break` in the code, so my initial idea would not work.
Still, I think I like this style more, but I'd like some second opinion on this. I would just remove these `getCompletedFieldsArray` and `takeWhile` functions, which you use only once, and write simply
```ts
const addressFields = fields.map(field => address[field.name])
const firstMissingIdx = addressFields.findIndex(field => !field)
const completedFields = firstMissingIdx === -1 
  ? addressFields 
  : addressFields.slice(0, firstMissngIdx)
```

react/LocationSelect.tsx
pedromtec: I prefer `const` because it makes the code easier to reason about. Since it can't be reassigned,  you will have less side effects or edge cases to reason about.

mredigonda: The source is [YDKJS](https://github.com/getify/You-Dont-Know-JS/blob/2nd-ed/get-started/ch2.md) (start reading from the first appearance of `const`).

I agree that avoiding reassignments is still good, and now that I'm thinking, maybe linters help with noticing mutations? 🤔 

Anyways, just to be extra clear with what kind of "undesirable" behavior you could get, here goes an example:

```javascript
const x = [1, 2, 3]
x[1] = 5
console.log(x)
```

Would show the array `[1, 5, 3]` while you would expect it to forbid you from making that change in the array.

react/LocationSelect.tsx
mredigonda: That seems awesome, thanks!

react/LocationSelect.tsx
marcoskwkm: Well, someone aware of the true meaning of `const` would not expect JS to forbid the array mutation =P
But I guess `let` is fine in this situation. It is the most common keyword in this construct anyway.

react/AddressForm.tsx
```diff
@@ -113,29 +114,25 @@ const AddressForm: StorefrontFunctionComponent = () => {
...
 
     const style = styleRules[field.label]
 
-    if (hasWithoutNumberOption(field.label))
-      return (
-        <div className="flex-auto mh3" style={style as React.CSSProperties}>
-          <NumberOption showCheckbox />
-        </div>
-      )
+    let fragmentComponent = undefined
```
marcoskwkm: We usually omit `= undefined` in this case.

react/AddressForm.tsx
```diff
@@ -108,30 +109,38 @@ const AddressForm: StorefrontFunctionComponent = () => {
...
-    )
+    ) {
       return null
+    }
 
-    if (hasWithoutNumberOption(field.label))
-      return <NumberOption showCheckbox />
+    const style = styleRules[field.label]
+
+    let fragmentComponent = undefined
```
lucasecdb: nit: rename this to `fragmentElement`, since it's won't be a component, but an element

react/LocationCountry.tsx
```diff
@@ -0,0 +1,93 @@
...
+const LocationCountry: StorefrontFunctionComponent<{}> = () => {
+  const intl = useIntl()
+  const { address, setAddress, countries } = useAddressContext()
+  const {
+    culture: { country: storeCountry },
+  } = useRuntime()
+  let { country } = address
+
+  if (country && countries && !countries.includes(country)) {
+    throw `The country ${country} doesn't belong to the country list, can't render LocationCountry`
```
lucasecdb: nit:
```suggestion
    throw `The country "${country}" doesn't belong to the country list, can't render LocationCountry`
```

react/LocationCountry.tsx
```diff
@@ -0,0 +1,93 @@
...
+        number: null,
+        postalCode: null,
+        state: null,
+        street: null,
+      }))
+    },
+    options,
+  }
+
+  return <Dropdown {...dropdownProps} />
```
lucasecdb: nit: I prefer the props to be inline in the element
```jsx
  return (
    <Dropdown
      label={<FormattedMessage id="place-components.label.country" />}
      value={country}
      placeholder="Select..."
      onChange={({ target }: React.ChangeEvent<HTMLSelectElement>) => {
        setAddress((prevAddress: Address) => ({
          ...prevAddress,
          city: null,
          complement: null,
          country: target.value,
          geoCoordinates: null,
          neighborhood: null,
          number: null,
          postalCode: null,
          state: null,
          street: null,
        }))
      }}
      options={options}
    />
  )
```

react/LocationSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+    id: 'place-components.label.locality',
+  },
+})
+
+const LocationSelect: StorefrontFunctionComponent<{}> = () => {
+  const { address, setAddress } = useAddressContext()
+  const countryRules = rules[address.country]
+
+  if (!countryRules.locationSelect) {
+    throw `The LocationSelect component is not applicable to this country: ${address.country}`
```
lucasecdb:
```suggestion
    throw `The country "${address.country}" is not applicable to the LocationSelect component`
```

react/LocationSelect.tsx
```diff
@@ -0,0 +1,111 @@
...
+    defaultMessage: '',
+    id: 'place-components.label.state',
+  },
+  locality: {
+    defaultMessage: '',
+    id: 'place-components.label.locality',
+  },
+})
+
+const LocationSelect: StorefrontFunctionComponent<{}> = () => {
```
lucasecdb: if you don't use the `schema` property, I'd rather have this declared as a `React.FC`

react/LocationSelect.tsx
mredigonda: > someone aware of the true meaning of const would not expect JS to forbid the array mutation

That's exactly what I thought 😅 Well, this code won't have `let` anyways because it has been rewritten with a more functional style (as we discussed in the other conversation) that also helps in avoiding mutations.

react/LocationSelect.tsx
mredigonda: I addressed all others nits but this one I would rather address in the latest code available for this project, since this is applicable for most of my components 👍

PR #7, opened 2020-02-13
Add device coordinates component

mredigonda:

This PR adds the DeviceCoordinates component, that should ask for permissions to get your location from the device you are using, and it should automatically fill data based on that location.

[Related clubhouse story](https://app.clubhouse.io/vtex/story/31851/criar-devicecoordinates)

[Workspace](https://devicecoordinates--checkoutio.myvtex.com/places)

This won't currently work since the `apiKey` in the code is not the real one, that shouldn't be disclosed because this code is open source. You can `vtex link` this locally (don't forget to also link `address-graphql` and `styleguide`) and place a valid `apiKey` to see how this works.

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.
- [x] Linked this PR to a Clubhouse story (if applicable).

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements


react/DeviceCoordinates.tsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  const onGetCurrentPositionSuccess = useCallback(
+    ({ coords }: Position) => {
+      console.log('Success!')
+      getAddressFromGeocoordinates({
+        variables: {
+          lat: coords.latitude.toString(),
+          lng: coords.longitude.toString(),
+          apiKey: 'PUT API KEY HERE',
```
mredigonda: Any ideas on this? This most likely will break running this in production, but it wouldn't be correct to place the real apiKey here.

darlenedms: It varies with accounts, right? So you should receive it from props https://github.com/vtex/address-form/blob/c08be681af8a352e53d37a8e3bc0ef831fa01c7a/demo/src/App.js#L85

mredigonda: Yes, I'll add it as a prop to the `address-context` component. But still, when you want to really test it, you should place the real apiKey in the code (after the change, it would be in the `Example.tsx` component).

react/DeviceCoordinates.tsx
```diff
@@ -0,0 +1,111 @@
...
+        enableHighAccuracy: true,
+      }
+    )
+  }, [onGetCurrentPositionSuccess, onGetCurrentPositionError])
+
+  useEffect(() => {
+    navigator.permissions
+      .query({ name: 'geolocation' })
+      .then((result: { state: string }) => {
+        if (result.state === 'granted') {
```
jeffersontuc: You could use a `const` instead of the string here.

react/DeviceCoordinates.tsx
```diff
@@ -0,0 +1,111 @@
...
+          onButtonClick()
+        } else if (result.state === 'denied') {
```
jeffersontuc: ditto

react/DeviceCoordinates.tsx
lucasecdb: these strings are from the Permissions API, and even the [MDN examples](https://developer.mozilla.org/en-US/docs/Web/API/Permissions#Example) show you to use them like this. would you still want this to be refactored to a constant?

jeffersontuc: It's ok for me!

react/AddressForm.tsx
```diff
@@ -77,7 +78,7 @@ const AddressForm: StorefrontFunctionComponent = () => {
   )
 
   const hasWithoutNumberOption = (label: string) => {
-    return label.length >= 6 && label.substr(label.length - 6) == 'Option'
+    return label.length >= 6 && label.substr(label.length - 6) === 'Option'
```
jeffersontuc: I would prefer to use a variable here saying what this magic number means.

lucasecdb: sure! but I don't know what number is this :joy:

jeffersontuc: :joy:

jeffersontuc: But considering how many open PRs this repo has, I'm not against merge it that way for now. In the future we gonna be able to understand what this number means.

lucasecdb: ok! I'm trying to fix some ts errors to merge this and move on to the next ones

marcoskwkm: > In the future we gonna be able to understand what this number means.

I should save this comment so in the future I can show you how wrong you were 🤣

react/__mocks__/vtex.styleguide.tsx
```diff
@@ -1,4 +1,10 @@
+// @ts-ignore
```
marcoskwkm: Why some lines need `@ts-ignore` and other don't? Do we actually need `@ts-ignore` here?

lucasecdb: yes because only some of the components don't have ts typings

react/AddressForm.tsx
mredigonda: HAHAHA well, when constructing an address you have various parts like "city", "street", "state", etc., and those get mapped depending on country to "fantasy names", like "province", "stateAbbreviation", "number", "numberOption".
The thing is, there were two fantasy names that had "Option" at the end, that had a very different functionality (a checkbox gets rendered that allows to disable the number input).
Howeeeeeeeeeeeeeeever, this could easily be done with Javascript's `.endsWith()` string function 🤦‍♂.

react/LocationCountry.tsx
```diff
@@ -44,7 +44,9 @@ const LocationCountry: StorefrontFunctionComponent<{}> = () => {
   let { country } = address
 
   if (country && countries && !countries.includes(country)) {
-    throw `The country "${country}" doesn't belong to the country list, can't render LocationCountry`
+    throw Error(
```
lucasecdb:
```suggestion
    throw new Error(
```

react/AddressForm.tsx
lucasecdb: thanks for the heads-up Maxi! 😄

PR #9, opened 2020-03-05
Add location input component

mredigonda:
Adds the `LocationInput` component.

[Related clubhouse story](https://app.clubhouse.io/vtex/story/32639/criar-locationfield-input)

[Workspace](https://lucas--checkoutio.myvtex.com/places)

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.
- [x] Linked this PR to a Clubhouse story (if applicable).

image

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements


react/DeviceCoordinates.tsx
```diff
@@ -77,27 +77,35 @@ const DeviceCoordinates: StorefrontFunctionComponent = () => {
...
-        icon = <IconLocation />
+        icon = <IconLocation block />
         break
       default:
     }
 
-    return loading ? <Spinner size={16} /> : icon
+    if (loading) {
+      icon = <Spinner size={16} />
+    }
```
marcoskwkm: Can't we set the state to `State.LOADING` when the query is loading? This `if` after the `switch` block seems bad.

react/LocationInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+      setAddress(data.getAddressFromPostalCode)
+      onSuccess?.(data.getAddressFromPostalCode)
+    }
+
+    if (error) {
+      console.warn(error.message)
+    }
+  }, [data, error, onSuccess, setAddress])
+
+  const onButtonClick = () => {
```
marcoskwkm: Shouldn't this be called `handleButtonClick`?

react/LocationInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <Input
+          label={<FormattedMessage id="place-components.label.postalCode" />}
+          button={button}
+          buttonProps={{ onClick: onButtonClick }}
+          value={inputValue}
+          onChange={({
+            target: { value },
+          }: React.ChangeEvent<HTMLInputElement>) => {
+            setInputValue(value)
+          }}
```
marcoskwkm: Maybe extract this to a `handleChange` function?

react/DeviceCoordinates.tsx
lucasecdb: sure!

react/DeviceCoordinates.tsx
```diff
@@ -1,107 +1,113 @@
...
+  const handleButtonClick = () => {
+    requestGeolocation()
+  }
 
   useEffect(() => {
     navigator.permissions
       .query({ name: 'geolocation' })
-      .then((result: { state: string }) => {
+      .then((result: PermissionStatus) => {
         if (result.state === 'granted') {
```
darlenedms: Why you didn't choose to use `PermissionState.GRANTED` here? Or another constant with those values.

react/LocationInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+    )
+  }
+
+  useEffect(() => {
+    if (data) {
+      setAddress(data.getAddressFromPostalCode)
+      onSuccess?.(data.getAddressFromPostalCode)
+    }
+
+    if (error) {
```
darlenedms: Why did you choose to not use the `if else` here?

react/DeviceCoordinates.tsx
lucasecdb: because `PermissionState` doesn't map to this value, as it is a value returned from the browser API. I could extract it to a constant, but I don't think it'd help much as the string is very descriptive

react/LocationInput.tsx
lucasecdb: no particular reason, I could change if you'd like

react/DeviceCoordinates.tsx
```diff
@@ -1,107 +1,113 @@
...
+        return <Spinner size={16} block />
+      case PermissionState.DENIED:
+        return <IconLocation block />
       default:
+        return null
     }
-
-    return loading ? <Spinner size={16} /> : icon
-  }
+  }, [geolocationPermission])
```
marcoskwkm: :satisfaction:

react/LocationInput.tsx
darlenedms: I was just curious because on `DeviceCoordinates.tsx` you (maybe Maxi, I don't know) have changed to use the `if else` and haven't done it here.
 ⛏ Would be nice to keep it consistent. I prefer this version, without the `else` but it's your call.

react/DeviceCoordinates.tsx
darlenedms: Understood.

react/LocationInput.tsx
lucasecdb: okay! I altered it to an `else if` but I also prefer the two independent `if`s 😅

PR #17, opened 2020-07-08
CHK-134: Input field of Location Search autocomplete component

wdsrocha:

Implements the input part of the following component.

image.

1. [Add any product to cart](https://search--checkoutio.myvtex.com/cart/add?sku=291);

### DIFF
--- a/react/LocationSearch.tsx
+++ b/react/LocationSearch.tsx
@@ -7,7 +7,12 @@ import styles from './LocationSearch.css'
 const LocationSearch: React.FC = () => {
   const [inputValue, setInputValue] = useState<string>('')
 
-  const onClear = () => setInputValue('')
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key !== 'Escape') {
+      return
+    }
+    setInputValue('')
+  }
 
   return (
     <div className={`${styles.locationSearch} w-100`}>
@@ -27,8 +32,8 @@ const LocationSearch: React.FC = () => {
               role="button"
               tabIndex={-1}
               className="pointer c-muted-3 flex justify-center-s items-center-m outline-0-m"
-              onClick={onClear}
-              onKeyPress={onClear}
+              onClick={() => setInputValue('')}
+              onKeyPress={() => {}}
             >
               <IconClear />
             </span>
@@ -38,6 +43,7 @@ const LocationSearch: React.FC = () => {
         onChange={(e: { target: { value: React.SetStateAction<string> } }) =>
           setInputValue(e.target.value)
         }
+        onKeyDown={handleKeyDown}
       />
     </div>
   )
--- a/react/LocationSearch.tsx
+++ b/react/LocationSearch.tsx
@@ -12,20 +12,25 @@ const LocationSearch: React.FC = () => {
   return (
     <div className={`${styles.locationSearch} w-100`}>
       <Input
+        testId="location-search-input"
         label={
           <FormattedMessage id="place-components.label.autocompleteAddress" />
         }
-        prefix={<IconSearch color="#134CD8" />}
+        prefix={
+          <div className="c-action-primary flex justify-center-s items-center-m">
+            <IconSearch />
+          </div>
+        }
         suffix={
           inputValue.length && (
             <span
               role="button"
               tabIndex={-1}
-              className="pointer"
+              className="pointer c-muted-3 flex justify-center-s items-center-m outline-0-m"
               onClick={onClear}
               onKeyPress={onClear}
             >
-              <IconClear color="#CACBCC" />
+              <IconClear />
             </span>
           )
         }
--- a/react/LocationSearch.tsx
+++ b/react/LocationSearch.tsx
@@ -2,13 +2,15 @@ import React, { useState } from 'react'
 import { Input, IconSearch, IconClear } from 'vtex.styleguide'
 import { FormattedMessage } from 'react-intl'
 
+import styles from './LocationSearch.css'
+
 const LocationSearch: React.FC = () => {
   const [inputValue, setInputValue] = useState<string>('')
 
   const onClear = () => setInputValue('')
 
   return (
-    <div className="w-100">
+    <div className={`${styles.locationSearch} w-100`}>
       <Input
         label={
           <FormattedMessage id="place-components.label.autocompleteAddress" />
@@ -18,7 +20,7 @@ const LocationSearch: React.FC = () => {
           inputValue.length && (
             <span
               role="button"
-              tabIndex={0}
+              tabIndex={-1}
               className="pointer"
               onClick={onClear}
               onKeyPress={onClear}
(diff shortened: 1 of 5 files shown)


react/LocationSearch.tsx
```diff
@@ -0,0 +1,41 @@
...
+        suffix={
+          inputValue.length && (
+            <span
+              role="button"
+              tabIndex={-1}
+              className="pointer"
+              onClick={onClear}
+              onKeyPress={onClear}
+            >
+              <IconClear color="#CACBCC" />
```
wdsrocha, line 28: How to use a tertiary color instead?

react/LocationSearch.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const onClear = () => setInputValue('')
+
+  return (
+    <div className={`${styles.locationSearch} w-100`}>
+      <Input
+        label={
+          <FormattedMessage id="place-components.label.autocompleteAddress" />
+        }
+        prefix={<IconSearch color="#134CD8" />}
```
wdsrocha, line 18: How to use a primary color instead?

wdsrocha:
On figma, this input have 40px/2.5rem of height, along with _other components_ that aren't respecting the design (such as postal code one). Should we adjust the height of this input and replicate to others? If yes, how to do this at CSS level? 👀

react/LocationSearch.tsx
```diff
@@ -0,0 +1,41 @@
...
+        suffix={
+          inputValue.length && (
+            <span
+              role="button"
+              tabIndex={-1}
```
lucasecdb, line 23: should be `tabIndex={0}`, else we will never be able to tab into this element
```suggestion
              tabIndex={0}
```

react/LocationSearch.tsx
```diff
@@ -0,0 +1,41 @@
...
+              className="pointer"
+              onClick={onClear}
+              onKeyPress={onClear}
```
lucasecdb, line 26: you need to check for the key in this event (either `enter` or `space`) to trigger the `onClear` function, else we could tab into this element and hit `tab` to go to the next tabbable element and end up clearing the input, which wouldn't be intended

react/LocationSearch.tsx
lucasecdb: you can use `className="c-action-primary"`

react/LocationSearch.tsx
lucasecdb: you can use `className="c-muted-3"`

lucasecdb:
> On figma, this input have 40px/2.5rem of height, along with _other components_ that aren't respecting the design (such as postal code one). Should we adjust the height of this input and replicate to others? If yes, how to do this at CSS level? 👀

IMO this is the responsibility of styleguide's `Input` component. If it does not adjust the height on mobile viewports we should fix it there

react/LocationSearch.tsx
wdsrocha: Do you really think we should be able to tab into it? As this behavior wasn't described in the design document, I just copied the behavior from other autocomplete components such as [VTEX Styleguide](https://styleguide.vtex.com/#/Components/Forms/AutocompleteInput), [Algolia](url) and [Material UI](https://material-ui.com/pt/components/autocomplete/).

lucasecdb: yes, if we don't make it tabbable we would degrade the experience for keyboard and screen reader users

lucasecdb: instead of making this tabbable, we could add a keydown listener on the input itself, and handle the `escape` key event as the clear., which is actually one of the specification requirements for a combobox (or autocomplete)

image

react/LocationSearch.tsx
wdsrocha: Only works when added on the span tag, but works very well 👌

react/LocationSearch.tsx
lucasecdb: there is [this example](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html) of a combobox with a listbox, which I think is the concept for the UI of this autocomplete, if you want to read more about accessibility

react/LocationSearch.tsx
wdsrocha: Only works when added on a wrapper div, and the div make it vertically unaligned. IMO adding this to the CSS will do no harm, but if there is a simpler way to do this, just tell me.

```
.locationSearch div :global(.vtex__icon-search) {
  display: flex;
  justify-content: center;
  align-items: center;
}
```

lucasecdb: can you add these styles to the div itself?

wdsrocha: Yes, I can (I forgot that I should be doing that lol, ty)

react/LocationSearch.tsx
wdsrocha: Didn't managed to make this work :(

I tried to use`onKeyDown` and `onKeyPressed` events, but even preventing the default behavior (as suggested by @marcoskwkm), I still can't capture `<Esc>`. On the other hand I can capture almost anything else.

I suppose the styleguide component is forcing some behavior under the hood, but I couldn't identify it on the[ input source code](https://github.com/vtex/styleguide/blob/master/react/components/Input/index.js).

Any ideas?

wdsrocha: Solved it! 🙃🙃

The problem: I use a Chrome extension that allows me to use Vim keybindings for navigation and it was capturing some keys before the component... that also explains why the components on w3 weren't behaving properly.

react/LocationSearch.tsx
```diff
@@ -0,0 +1,52 @@
...
+              className="pointer c-muted-3 flex justify-center-s items-center-m outline-0-m"
+              onClick={() => setInputValue('')}
+              onKeyPress={() => {}}
+            >
+              <IconClear />
+            </span>
+          )
+        }
+        value={inputValue}
+        onChange={(e: { target: { value: React.SetStateAction<string> } }) =>
```
lucasecdb, line 43: nitpick: why not?
```suggestion
        onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
```

react/LocationSearch.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <div className="c-action-primary flex justify-center-s items-center-m">
+            <IconSearch />
+          </div>
+        }
+        suffix={
+          inputValue.length && (
+            <span
+              role="button"
+              tabIndex={-1}
+              className="pointer c-muted-3 flex justify-center-s items-center-m outline-0-m"
```
lucasecdb, line 34: why do you need the `-s` suffix in the `justify-center` class? same as for the `-m` on `items-center` and `outline-0`

react/LocationSearch.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <div className={`${styles.locationSearch} w-100`}>
+      <Input
+        testId="location-search-input"
+        label={
+          <FormattedMessage id="place-components.label.autocompleteAddress" />
+        }
+        prefix={
+          <div className="c-action-primary flex justify-center-s items-center-m">
```
lucasecdb, line 25: same as above

wdsrocha:
> This doesn't implement the autocomplete itself yet, just the input, right? Just to be sure.

Yep.

PR #20, opened 2020-07-15
CHK-144: Drop-down field logic of Location Search autocomplete component 

wdsrocha:

Continuation of #17, implements the logical part (no styling) of the following component.

image.

1. [Add any product to cart](https://wdsrocha--checkoutio.myvtex.com/cart/add?sku=291);

#### Notes

I'm not sure _how much_ will need to change when integrating this with the real queries, but I don't think it really matters right now (as this is not even going to master). Also, `react/addresses.tsx` will be deleted later on.


react/LocationSearch.tsx
```diff
@@ -1,50 +1,92 @@
...
+  ComboboxList,
+} from '@reach/combobox'
+import '@reach/combobox/styles.css'
 
+import { addresses as mockedAddresses } from './addresses'
 import styles from './LocationSearch.css'
 
-const LocationSearch: React.FC = () => {
-  const [inputValue, setInputValue] = useState<string>('')
+const useAddressMatch = (term: string): string[] => {
```
lucasecdb, line 16: this isn't a hook, is it?

react/LocationSearch.tsx
```diff
@@ -1,50 +1,92 @@
...
+  return mockedAddresses
+    .filter((address: string) =>
+      address.toLocaleLowerCase().includes(term.toLocaleLowerCase())
+    )
+    .slice(0, 6)
```
lucasecdb, line 21: why these arguments? can you extract them to a constant or add a comment?

react/LocationSearch.tsx
wdsrocha: It isn't a hook, but now I guess it should be an effect hook, right?

react/LocationSearch.tsx
wdsrocha: didn't put too much thought into it, as I supposed that this should be handled on the GraphQL query. Will add a comment

react/LocationSearch.tsx
lucasecdb: hmm not necessarily... maybe a memo, but I don't think it would make much of a difference here. might be better to define it as a plain function and if we run into trouble later we could convert it to a hook

react/LocationSearch.tsx
lucasecdb: I got confused with this (read too fast haha), but this is to limit the number of addresses shown in the dropdown right? we could convert the 6 to a constant like `MAX_DROPDOWN_ADDRESSES` or something

react/LocationSearch.tsx
wdsrocha: That means literally a function like `getAddresses` and run it when the input changes?

react/LocationSearch.tsx
wdsrocha: Yes. And sure, the constant will definitely make it easier to read, thks.

react/LocationSearch.tsx
```diff
@@ -1,50 +1,95 @@
...
+  }
+
+  const handleAddressSelection = (selectedAddress: string) => {
+    setSearchTerm(selectedAddress)
+    onSelectAddress?.(selectedAddress)
+  }
+
+  // This function will be replaced in the future, after integrating the
+  // component with GraphQL queries.
+  const getAddresses = () => {
```
marcoskwkm, line 39: I know this is temporary, but this is an expensive function so you should memoize its return value. I think the way it was before was good, you just needed to add an `useMemo`.
```ts
const addresses = useMemo(() => getAddresses(searchTerm), [searchTerm])
```

marcoskwkm:
The autocomplete doesn't seem to be working for me. I tried typing substrings of some mocked addresses and nothing popped up.

<p align="center"></p>

wdsrocha:
> The autocomplete doesn't seem to be working for me. I tried typing substrings of some mocked addresses and nothing popped up.

Whoops, I broke it with the last commit :harold-pain:. Will fix in the next one.

wdsrocha:
> DId you change the mocked addresses list when you linked? The autocomplete is working now but it doesn't find most of the addresses (it seems to only find addresses starting with 'a' or 'b'). It also doesn't show the address number.

Yes, my bad. I'll keep in mind to change the workspace when working in another branch from now on. You may take a look now, @marcoskwkm, I'll let this linked correctly until the PR is merged.

Btw the change to the mocked addresses was just a test and I don't think I'll change it in the next PR or something.

PR #21, opened 2020-07-16
Refactor and style Location Search autocomplete

wdsrocha:

Continuation of #17 and #20, implements the styling part of the following component.

image

1. [Add any product to cart](https://combobox--checkoutio.myvtex.com/cart/add?sku=291);

#### Usage example

```typescript
import GoogleLogo from './../assets/powered_by_google.png'

// ...

<LocationSearch
  renderEngineLogo={() => (
    <img className="h1" src={GoogleLogo} alt="Powered by Google" />
  )}
/>
```


react/LocationSearch.tsx
```diff
@@ -48,46 +55,69 @@ const LocationSearch: React.FC<LocationSearchProps> = ({ onSelectAddress }) => {
...
+                  <IconClear />
+                </span>
+              )
+            }
+            value={searchTerm}
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+              setSearchTerm(event.target.value)
+            }
+            onKeyDown={handleKeyDown}
+          />
```
wdsrocha, line 89: No "real" changes here, just wrapped the `ComboboxInput` inside a `div`.

react/addresses.tsx
```diff
@@ -1,4 +1,6 @@
 export const addresses: string[] = [
+  'Lorem ipsum dolor sit amet consectetur adipisicing elit. Nihil itaque repellendus adipisci ex ipsum beatae iure. Tenetur ea, veritatis libero unde consectetur eaque sunt debitis soluta reiciendis non. Ad, non.',
+  'The quick brown fox jumps over the lazy dog',
```
wdsrocha, line 3: This is helpful when testing text overflow.

react/LocationSearch.tsx
```diff
@@ -27,11 +27,18 @@ const getAddresses = (searchTerm: string) => {
...
 }
 
-const LocationSearch: React.FC<LocationSearchProps> = ({ onSelectAddress }) => {
+const LocationSearch: React.FC<LocationSearchProps> = ({
+  onSelectAddress,
+  renderEngineLogo,
+}) => {
   const [searchTerm, setSearchTerm] = useState<string>('')
   const addresses = useMemo(() => getAddresses(searchTerm), [searchTerm])
+  const inputWrapperEl = useRef<HTMLDivElement>(null)
```
lucasecdb, line 41: nit: we usually name these ref variables with a `Ref` suffix
```suggestion
  const inputWrapperRef = useRef<HTMLDivElement>(null)
```

react/LocationSearch.tsx
```diff
@@ -48,46 +55,69 @@ const LocationSearch: React.FC<LocationSearchProps> = ({ onSelectAddress }) => {
...
-                data-testid="location-search-clear"
-                role="button"
-                tabIndex={-1}
-                className="pointer c-muted-3 flex justify-center items-center outline-0"
-                onClick={() => setSearchTerm('')}
-                onKeyPress={() => {}}
-              >
-                <IconClear />
-              </span>
+        <div ref={inputWrapperEl}>
```
lucasecdb, line 58: related to the comment above ☝️ 

```suggestion
        <div ref={inputWrapperRef}>
```

react/LocationSearch.tsx
```diff
@@ -48,46 +55,69 @@ const LocationSearch: React.FC<LocationSearchProps> = ({ onSelectAddress }) => {
...
+              <div className="c-action-primary flex justify-center items-center">
+                <IconSearch />
+              </div>
+            }
+            suffix={
+              searchTerm.trim().length && (
+                <span
+                  data-testid="location-search-clear"
+                  role="button"
+                  tabIndex={-1}
```
lucasecdb, line 75: do you think it is worth it to comment why we add the `tabIndex={-1}` or nah?

react/LocationSearch.tsx
```diff
@@ -27,11 +27,18 @@ const getAddresses = (searchTerm: string) => {
 
 interface LocationSearchProps {
   onSelectAddress?: (selectedAddress: string) => void
+  renderEngineLogo?: (
+    props: React.ImgHTMLAttributes<HTMLImageElement> & { className?: string }
+  ) => HTMLImageElement
```
lucasecdb, line 32: hmmm, I don't know how I feel about this parameter and return value.. I think it's best to let the user of this component decide what to render here, no? in case we want to render a svg instead of a img for example

react/components/Combobox.tsx
```diff
@@ -0,0 +1,94 @@
...
+    {...props}
+    className={classNames(
+      className,
+      styles.option,
+      'flex items-center pointer bg-action-secondary pv3 ph5 hover-bg-action-secondary'
+    )}
+  >
+    {children ?? (
+      <>
+        <Icon className="flex-shrink-0 mr4" />
```
lucasecdb, line 79: wdyt of having an `icon` prop here instead of hard-coding the search icon? it would make it easier to extract this component to another repo or to styleguide itself

react/LocationSearch.tsx
wdsrocha: I can add something like "the input can be cleared by pressing the `esc` key, so the clear button should not be tabbable". Do you think that will make things clear?

wdsrocha: No pun intended :))

lucasecdb: yeah haha I think it should clear up any concerns regarding accessibility with this component

react/LocationSearch.tsx
wdsrocha: Agreed, good point. I'm just unsure on how to do that. Does `renderEngineLogo?: (props: React.HTMLAttributes<HTMLElement>) => HTMLElement` makes sense? 🤔

react/LocationSearch.tsx
```diff
@@ -100,7 +101,11 @@ const LocationSearch: React.FC<LocationSearchProps> = ({
             <>
               <ComboboxList>
                 {addresses.map((address, index) => (
-                  <ComboboxOption value={address} key={index} />
+                  <ComboboxOption
+                    renderIcon={() => <PlaceIcon />}
```
wdsrocha, line 105: idk, I have mixed feelings about this. @lucasecdb when we discussed, I imagined this implementation as `icon={PlaceIcon}`, but:
- somehow it wasn't working;
- as I don't really see the benefit by picking one or the other (in that case), I decided to render it like that.

lucasecdb: I was thinking more on a `icon={<PlaceIcon />}` to be honest..

davicosta99:
- The full address is not being shown. In the native Google component, and in the Figma file, the neighborhood and other secondary info is show in-line after the street name and number. [Example frame](https://www.figma.com/file/umwHrHA8nifvQIPEN3DHpX/IO-Styleguide?node-id=97%3A16).
- The icon before each suggestion should use the color from muted-1 token (#727273). Something happened in the Figma file, but I've fixed it. [Example frame](https://www.figma.com/file/umwHrHA8nifvQIPEN3DHpX/IO-Styleguide?node-id=97%3A16).
- When a suggestion is selected (on:active) it should blink in the on-action color token (#D2DEFC) for 50ms before closing.
- The max-width for the field is not the same as in the Figma file. [Example frame](https://www.figma.com/file/umwHrHA8nifvQIPEN3DHpX/IO-Styleguide?node-id=97%3A16)).
- When hovered the clear icon should use the color from muted-2 token (#979899).

react/LocationSearch.tsx
```diff
@@ -1,37 +1,60 @@
...
+  mainText: string
+  mainTextMatchInterval: Interval
+  secondaryText: string
+}
+
+const renderSuggestionText = (suggestion: Suggestion) => {
+  const { mainText } = suggestion
+  const { offset, size } = suggestion.mainTextMatchInterval
+  return (
+    <div className="truncate c-muted-2">
```
wdsrocha, line 32: When the text inside an Option is too big, it will truncate with an ellipsis, but there is a problem: the `mainText` (street name) is black and the `secondaryText` (city and state name) is gray, which would require the ellipsis color to be dynamic, depending on which part have overflowed.

Doing this at implementation level is quite painful and don't seems to bring too much value (big street names aren't very common, I guess), so I've opted to always display the ellipsis with a gray color. Is that okay? @davicosta99 @augustocb

Example of fake big street name with the grey colored ellipsis:
image

react/components/Combobox.css
```diff
@@ -0,0 +1,15 @@
...
+  --reach-combobox: 1;
+}
+
+.option:not([aria-selected="true"]).option:not(:hover) {
+  background-color: transparent;
+}
+
+.option:active {
+  background-color: #d2defc
+}
```
wdsrocha, line 11: @davicosta99, while addressing the following issue

> When a suggestion is selected (on:active) it should blink in the on-action color token (#D2DEFC) for 50ms before closing.

I found out that this color does not exist on [VTEX Tachyons spec](https://github.com/vtex/vtex-tachyons/blob/master/tachyons.css), so I’ve added it directly here.

Also, how do I know if it takes 50ms during the blink? Is that the standard behavior?

wdsrocha: Same goes for `#c4c4c4` on Place Icon and `#2e2e2e` on Warning Icon, but I used the closest "c-muted-x" for them:
- `#c4c4c4` -> `#cacbcc` (`c-muted-3`)
- `#2e2e2e` -> `#727273` (`c-muted-1`)

wdsrocha:
@davicosta99 

> The max-width for the field is not the same as in the Figma file.

This task/PR only addresses the `LocationSearch` component itself, so I'll worry with the resize later on, when integrating with the `NewFormAddress` component inside the shipping step. But thanks for pointing that out, I ended up finding a bug related to that!

> I'm not able to test on my phone because after submitting the profile step I get this error

I don't know why this is happening to you, just tried on my phone and it seems to be working fine. Could you try again, starting from the [cart step](https://combobox--checkoutio.myvtex.com/cart/add?sku=291)?

react/LocationSearch.tsx
```diff
@@ -1,37 +1,59 @@
...
+      <span className="c-on-base">{mainText.substr(0, offset)}</span>
+      <span className="c-on-base b">{mainText.substr(offset, size)}</span>
+      <span className="c-on-base">{mainText.substr(size + offset)}</span>
+      <span>{` ${suggestion.secondaryText}`}</span>
+    </div>
+  )
 }
```
wdsrocha, line 38: #### About the full address being displayed

In order to handle this, I’ve opted for implementing something similar to Google autocomplete ([see AutocompletePrediction Interface](https://developers.google.com/maps/documentation/javascript/reference/places-autocomplete-service#AutocompletePrediction)). When the user inputs the text "`bri`" into the input box, an answer like this will be returned:

```json
[
  {
    "description": "Cambridge, Rockford, Illinois",
    "mainText": "Cambridge",
    "secondaryText": "Rockford, Illinois",
    "mainTextMatchInterval": {
      "offset": 3,
      "size": 3
    }
  },
  {
    "description": "Briar Crest, Hagerstown, Maryland",
    "mainText": "Briar Crest",
    "secondaryText": "Hagerstown, Maryland",
    "mainTextMatchInterval": {
      "offset": 0,
      "size": 3
    }
  }
]
```

Which will be converted to (just the first one, for simplicity):

```html
<div className="truncate c-muted-2">
  <span className="c-on-base">Cam</span>
  <span className="c-on-base b">bri</span>
  <span className="c-on-base">dge</span>
  <span>Rockford, Illinois</span>
</div>
```

And displayed like:



Explanation of the fields:

- `description`: the text that will go to the input box when the user selects it
- `mainText`: probably the street name (along with the number, maybe?)
- `secondaryText`: larger administrative regions like city, state and country
- `mainTextMatchInterval`: used to highlight the part of the `mainText` that matched the user input. This can be improved to hold multiple matches and/or include the `secondaryText`, but for now, I think imitating the Google behavior might be sufficient (which is just like that).

This isn't very flexible, but looks sufficient to me. Opinions?

react/LocationSearch.tsx
davicosta99: I think it's fine, Wesley, thanks for the concern. Appreciate the level of care 😉

react/components/Combobox.css
davicosta99: Currently we have these tokens, that basically reuse the hover color.
.active-bg-action-secondary:active { background-color: #dbe9fd; }
.active-b--action-secondary:active { border-color: #dbe9fd; }

Can you change them to #D2DEFC? This change was made in Figma but never implemented.

Regarding the 50ms, I thought this was something that needed to be defined. Right now is perfect already.

Sorry for the mistake on the grays, my bad. Thanks for using the right tokens 😉

react/LocationSearch.tsx
davicosta99: Seems enough for me as well.

davicosta99:
Besides including the color in our Tachyons like I've mentioned above, the only other issue I found was this:

Screen-Recording-2020-07-29-at-16 34 12

It seems like when the secondary text is included in the search term, no result is found. Is this a result of the mocked data or does something need to be corrected?

react/LocationSearch.tsx
```diff
@@ -1,37 +1,59 @@
...
-    )
-    .slice(0, MAX_DROPDOWN_ADDRESSES)
+const renderSuggestionText = (suggestion: Suggestion) => {
+  const { mainText } = suggestion
+  const { offset, size } = suggestion.mainTextMatchInterval
+  return (
+    <div className="truncate c-muted-2">
+      <span className="c-on-base">{mainText.substr(0, offset)}</span>
+      <span className="c-on-base b">{mainText.substr(offset, size)}</span>
+      <span className="c-on-base">{mainText.substr(size + offset)}</span>
```
lucasecdb, line 34: can we opt in to a simpler markup? like the following
```suggestion
      <span className="c-on-base">
        {mainText.substr(0, offset)}
        <span className="b">{mainText.substr(offset, size)}</span>
        {mainText.substr(size + offset)}
      </span>
```

wdsrocha: I think so, it looks better indeed. I was just worried that this wouldn't be good at a semantic level, but thinking again, I guess there is nothing wrong with a nested `span`.

lucasecdb: yep, it's an inline element inside another inline element, so I think we should be fine

wdsrocha:
> It seems like no result is found when the secondary text is included in the search term. Is this happening because of the mocked data or does something need to be corrected?

@davicosta99 as commented [here](https://vtex.slack.com/archives/C98FX9724/p1596052478252000?thread_ts=1595432482.014300&cid=C98FX9724), the current search algorithm is pretty naive and for testing purposes only. But I'll keep an eye on caching, as I suppose it does not make too much sense to re-do a request for this kind of search that you have pointed out.

react/LocationSearch.tsx
```diff
@@ -1,37 +1,61 @@
...
+  const { mainText } = suggestion
+  const { offset, size } = suggestion.mainTextMatchInterval
+  return (
+    <div className="truncate c-muted-2">
+      <span className="c-on-base">
+        {mainText.substr(0, offset)}
+        <span className="b">{mainText.substr(offset, size)}</span>
+        {mainText.substr(size + offset)}
+      </span>
+      <span>{` ${suggestion.secondaryText}`}</span>
```
lucasecdb, line 37: sorry I missed this in the last review, but it's just a nitpick
```suggestion
      {' '}<span>{suggestion.secondaryText}</span>
```

wdsrocha: Done, but the formatter sent the whitespace to the line above.

react/components/Combobox.tsx
```diff
@@ -55,7 +55,7 @@ export const ComboboxOption: React.FC<ReachComboboxOptionProps & {
     className={classNames(
       className,
       styles.option,
-      'flex items-center pointer bg-action-secondary pv3 ph5 hover-bg-action-secondary'
+      'flex items-center pv3 ph5 pointer bg-action-secondary hover-bg-action-secondary active-bg-action-secondary'
```
wdsrocha, line 58: Now this is "incorrect", but as soon as Styleguide bumps the Tachyons version, I'll open a PR bumping the Styleguide version here and it will have the expected color.

lucasecdb: you don't actually need to bump styleguide here after deploying the new version with `vtex-tachyons` updated, because the IO infra will automatically bump the styleguide version on the account itself when it is published.

PR #23, opened 2020-08-05
CHK-247: Fix unit test environment

wdsrocha:

The unit tests weren't working when running a `yarn test`. Although almost all tests are still broken, now we are _able to_ fix them.

1. Open this branch with `git fetch origin fix/tests && git switch fix/tests`
2. Run `cd react && yarn && yarn test`
3. `4` tests out of `4` should fail

#### How does this PR make you feel?

_When they ask if the tests are now working_


wdsrocha:
@lucasecdb I realized that many fixtures weren't being used, and after removing them, there wasn't many fixtures to justify the existence of three files. In the other hand, it seems silly to have a single fixture file inside a fixture directory. What do you think?

lucasecdb:
I think it's fine even if it's just one file

react/package.json
```diff
@@ -3,6 +3,8 @@
     "test": "vtex-test-tools test"
   },
   "dependencies": {
+    "apollo-cache-inmemory": "^1.6.6",
+    "apollo-client": "^2.6.10",
```
lucasecdb, line 7: shouldn't this go in `devDependencies`?

PR #24, opened 2020-08-05
CHK-152: Add tests and fix popover bug

wdsrocha:

Add few tests to the new component and fixes a bug where the popover would be open even though there were no input at all (that was raising a warning from React Testing Library, as the popover wasn't closed when the test ended).

1. Open this branch with `git fetch origin test/location-search && git switch test/location-search`
2. Run `cd react && yarn && yarn test __tests__/LocationSearch.test.tsx`
3. `3` tests out of `3` should pass

#### Notes

A warning is raised when running the tests:

```
$ vtex-test-tools test __tests__/LocationSearch.test.tsx
  console.error node_modules/@vtex/test-tools/node_modules/react/cjs/react.development.js:167
    Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?
```

How to fix that?

### DIFF
--- a/react/__tests__/LocationSearch.test.tsx
+++ b/react/__tests__/LocationSearch.test.tsx
@@ -1,0 +1,64 @@
+import React from 'react'
+import { render, screen, fireEvent } from '@vtex/test-tools/react'
+
+import LocationSearch from '../LocationSearch'
+import { getFewAddresses } from '../__fixtures__/fixtures'
+
+const UNTABABBLE = -1
+
+describe('Location Search', () => {
+  it('should be able to select a suggested address when a search term is found', async () => {
+    render(<LocationSearch getAddresses={getFewAddresses} />)
+
+    const input = screen.getByLabelText(
+      /Address and house number/i
+    ) as HTMLInputElement
+    fireEvent.change(input, { target: { value: 'Praia de Botafogo' } })
+
+    expect(
+      await screen.findByText('Praia de Botafogo, 200')
+    ).toBeInTheDocument()
+    expect(
+      await screen.findByText('Praia de Botafogo, 300')
+    ).toBeInTheDocument()
+
+    fireEvent.click(screen.getByText('Praia de Botafogo, 200'))
+    expect(input).toHaveValue(
+      'Praia de Botafogo, 200, Botafogo, Rio de Janeiro, RJ'
+    )
+  })
+
+  it('should be able to clear the input when something is typed', () => {
+    render(<LocationSearch getAddresses={getFewAddresses} />)
+
+    expect(screen.queryByRole('button')).not.toBeInTheDocument()
+
+    const input = screen.getByLabelText(
+      /Address and house number/i
+    ) as HTMLInputElement
+    fireEvent.change(input, { target: { value: 'Praia de Botafogo' } })
+    const button = screen.getByRole('button')
+    expect(button.tabIndex).toBe(UNTABABBLE)
+
+    fireEvent.click(button)
+
+    expect(input).toHaveValue('')
+    expect(screen.queryByRole('button')).not.toBeInTheDocument()
+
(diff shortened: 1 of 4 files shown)


lucasecdb:
> A warning is raised when running the tests:
> 
> ```
> $ vtex-test-tools test __tests__/LocationSearch.test.tsx
>   console.error node_modules/@vtex/test-tools/node_modules/react/cjs/react.development.js:167
>     Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?
> ```
> 
> How to fix that?

I think that is an issue on some styleguide component, and probably to fix it would be to correctly assign the `propTypes` or `defaultProps` on the component returned by `forwardRef` there, not on the function you pass as an argument

react/__tests__/LocationSearch.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { getFewAddresses } from '../__mocks__/LocationSearchMock'
+
+const UNTABABBLE = -1
+
+describe('Location Search', () => {
+  it('should be able to select a suggested address when a search term is found', async () => {
+    render(<LocationSearch getAddresses={getFewAddresses} />)
+
+    const input = screen.getByLabelText(
+      'Address and house number'
```
lucasecdb, line 14: nitpick comment: you can use regex with the `i` flag so the message can be case-insensitive

react/__tests__/LocationSearch.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+      await screen.findByText('Praia de Botafogo, 200')
+    ).toBeInTheDocument()
+    expect(
+      await screen.findByText('Praia de Botafogo, 300')
+    ).toBeInTheDocument()
+
+    fireEvent.click(screen.getByText('Praia de Botafogo, 200'))
+    expect(input.value).toBe(
+      'Praia de Botafogo, 200, Botafogo, Rio de Janeiro, RJ'
+    )
```
lucasecdb, line 28: nit
```suggestion
    expect(input).toHaveValue(
      'Praia de Botafogo, 200, Botafogo, Rio de Janeiro, RJ'
    )
```

react/__tests__/LocationSearch.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const input = screen.getByLabelText(
+      'Address and house number'
+    ) as HTMLInputElement
+    fireEvent.change(input, { target: { value: 'Praia de Botafogo' } })
+    const button = screen.getByRole('button')
+    expect(button.tabIndex).toBe(UNTABABBLE)
+
+    fireEvent.click(button)
+    expect(input.value).toBe('')
```
lucasecdb, line 44: nit
```suggestion
    expect(input).toHaveValue('')
```

react/__tests__/LocationSearch.test.tsx
wdsrocha: Why do you think this is better? Because of the failure message?

react/__tests__/LocationSearch.test.tsx
wdsrocha: I prefer keeping it that way because

- it's simpler
- it would fail the test if someone changed the case to something like "Address and House Number"

unless this is a common/best practice and I'm not aware of?

lucasecdb: but _should_ the test fail only because the casing changed? the end user isn't impacted at all if the casing is "Address and House Number" or "Address and house number", that's why I said about using the case insensitive regex

react/__tests__/LocationSearch.test.tsx
lucasecdb: yeah, I'd rather just add the expect on the `input` element and use the matchers on `jest-dom` if I can. anyway, this is only a nitpick

PR #29, opened 2020-11-10
CHK-138: Add label to CountryLocation

wdsrocha:

Label wasn't showing up in LocationCountry component.

- [Workspace](https://w0geo--checkoutio.myvtex.com/cart/add?sku=289)
- Go to shipping step
- Compare to [Figma](https://www.figma.com/file/umwHrHA8nifvQIPEN3DHpX/Checkout-IO---Library?node-id=97%3A15)

|Figma|Proposal|
|-|-|
| image | image |

#### Notes

The LocationCountry component will appear on a 50% chance _for this workspace only_. The reason for that is because the condition for LocationCountry appear is to have multiple countries defined by the sales channels. There is no easy way to simulate that on IO yet, and making two separate workspaces isn't possible due to a very recent store framework (?) bug. So please, refresh the page until it works :worry-sparkle:

### DIFF
--- a/react/LocationCountry.tsx
+++ b/react/LocationCountry.tsx
@@ -50,7 +50,13 @@ const renderCountryFlagWithName = ({
   </>
 )
 
-const LocationCountry: React.FC = () => {
+interface LocationCountryProps {
+  labelId?: string
+}
+
+const LocationCountry: React.FC<LocationCountryProps> = ({
+  labelId = 'location-country',
+}) => {
   const intl = useIntl()
   const { address, setAddress, countries } = useAddressContext()
   const {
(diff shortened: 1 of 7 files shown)


react/LocationCountry.tsx
```diff
@@ -51,6 +51,7 @@ const renderCountryFlagWithName = ({
 )
 
 const LocationCountry: React.FC = () => {
+  const labelId = 'location-country'
```
lucasecdb, line 54: this could be a prop if you'd ever want to change this label

wdsrocha: sure

lucasecdb (approved):

lucasecdb (approved):

jeffersontuc (approved):

react/LocationSearch.tsx
```diff
@@ -147,12 +142,13 @@ const LocationSearch: React.FC<LocationSearchProps> = ({
       address => address.description === selectedAddress
     )?.externalId
 
-    if (id) {
-      setSearchTerm(selectedAddress)
-      executeGetAddressByExternalId({ variables: { id } })
-    } else {
+    if (id === undefined) {
```
lucasecdb, line 145: why not

```suggestion
    if (id == null) {
```

wdsrocha: Hmm, linter told me that id could be undefined, so I just googled how to check for undefined and found this: https://stackoverflow.com/a/3390468/7651928

So, why do you think it's a good idea to compare it to null? Or why not just check for falsy values (in this case seems to be ok)?

```suggestion
    if (!id) {
```

lucasecdb: I had suggested comparing to null (using `==` instead of `===`) because it checks for both `null` and `undefined`.

```js
console.log(null == undefined) // true
```

lucasecdb (commented):
can you rebase?

wdsrocha:
Sure, I rebased and forgot to push :msn-embarrassed:

react/LocationInput.tsx
```diff
@@ -36,7 +41,7 @@ const LocationInput: React.FC<Props> = ({
   useEffect(() => {
     let cancelled = false
 
-    if (data) {
+    if (data?.getAddressFromPostalCode) {
```
wdsrocha, line 44: @lucasecdb looks ok to you? Just making sure that you see that I've refactored LocationInput (basically just added query typings)

lucasecdb: yep! lgtm

lucasecdb (approved):

PR #30, opened 2020-11-11
CHK-441: Add line height and remove icon in edit button

lucasecdb:

Conformity with our design.

[Workspace](https://billing--checkoutio.myvtex.com/cart/add?sku=289).

Note that the above workspace is still WIP, but you can see the updated component in the billing address inside the payment step.

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

Before:
image
<p align="center"><i>Editing only missing fields from autocomplete</i></p>

image
<p align="center"><i>Editing all fields, notice the pencil icon next to the postal code</i></p>

After:
image
<p align="center"><i>Editing only missing fields from autocomplete</i></p>

image
<p align="center"><i>Editing all fields</i></p>

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
_ | New feature 
_ | Breaking change 
✔️ | Technical improvements

### DIFF
--- a/react/PlaceDetails.tsx
+++ b/react/PlaceDetails.tsx
@@ -2,18 +2,31 @@ import msk from 'msk'
 import React from 'react'
 import { useAddressContext } from 'vtex.address-context/AddressContext'
 import { Display, AddressFields } from 'vtex.address-context/types'
+import { ButtonPlain } from 'vtex.styleguide'
+import { useIntl, defineMessages } from 'react-intl'
 
 interface Props {
   display?: keyof Display
   hiddenFields?: AddressFields[]
+  onEdit?: () => void
 }
+
+const messages = defineMessages({
+  edit: {
+    defaultMessage: '',
+    id: 'place-components.label.edit',
+  },
+})
 
 const PlaceDetails: React.FC<Props> = ({
   display = 'extended',
   hiddenFields = [],
+  onEdit,
 }) => {
   const { address, rules } = useAddressContext()
   const countryRules = rules[address.country as string]
+
+  const intl = useIntl()
 
   if (!countryRules) {
     return null
@@ -24,7 +37,7 @@ const PlaceDetails: React.FC<Props> = ({
   return (
     <div className="flex flex-column">
       {displaySpec.map((line, displayIndex) => (
-        <div key={displayIndex}>
+        <div className="flex" key={displayIndex}>
           {line.map((fragment, index) => {
             if (
               !address[fragment.name] ||
@@ -50,7 +63,7 @@ const PlaceDetails: React.FC<Props> = ({
               : address[fragment.name]
 
             return (
-              <span key={fragment.name}>
+              <div className="dib" key={fragment.name}>
                 {fragment.delimiter && hasPreviousFragment && (
                   <span>{fragment.delimiter}</span>
                 )}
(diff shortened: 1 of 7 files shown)


wdsrocha (approved):
It's that "editar" button, right? Just making sure I'm looking at the right thing. If that's correct, LGTM 👌 

image

lucasecdb:
> It's that "editar" button, right? Just making sure I'm looking at the right thing. If that's correct, LGTM 👌

it is not 😅 it should be on the right side of the street, but I've encountered an issue:

image

the gap between the street name and the edit button is because we have one whole div that contains the `PlaceDetails` component, and the edit button is outside of it. I'm thinking of adding a prop to `PlaceDetails` to allow us to render something inside of it, what do you think? I'm just not sure on the naming/usage of this prop yet

wdsrocha:
Are you sure the gap isn't intentional? Looking at the [figma design](https://www.figma.com/file/VXANxSsL0czYhnZi64mvto/Official-Documentation?node-id=126%3A0) it does not help much, as the street name has the biggest width among other address components, so maybe we should ask designers how it should look when the street name is smaller?

wdsrocha (changes requested):
Just undoing my wrong review :msn-embarrassed:

lucasecdb:
> Are you sure the gap isn't intentional? Looking at the [figma design](https://www.figma.com/file/VXANxSsL0czYhnZi64mvto/Official-Documentation?node-id=126%3A0) it does not help much, as the street name has the biggest width among other address components, so maybe we should ask designers how it should look when the street name is smaller?

good point! will ask

lucasecdb:
@wdsrocha fixed it in 4bb6a5c, can you take a look? feel free to suggest any changes in the component API as well
image

react/PlaceDetails.tsx
```diff
@@ -50,17 +63,24 @@ const PlaceDetails: React.FC<Props> = ({
...
                 )}
                 <span className={fragment.name}>{addressValue}</span>
                 {fragment.delimiterAfter && shouldShowDelimiter && (
                   <span>{fragment.delimiterAfter}</span>
                 )}
-              </span>
+              </div>
             )
           })}
+          {onEdit && displayIndex === 0 && (
```
wdsrocha, line 77: If `!address[fragment.name] || hiddenFields.includes(fragment.name)` is true, the edit button will appear alone with it's left margin:

print
|-
|

Can you verify how this case should be handled?

lucasecdb: sure

lucasecdb: fixed on 2e47258

wdsrocha: Nice!! I'll review soon (or ping me when you finish/stop working on the billing workspace)

wdsrocha (approved):
Noice

react/formatAddressToString.ts
```diff
@@ -0,0 +1,43 @@
...
+) {
+  const displaySpec = rules.display.compact
+
+  return displaySpec
+    .map(line => line.filter(fragment => address[fragment.name]))
+    .filter(line => line.length > 0)
+    .map(line =>
+      line
+        .map((fragment, index) => {
+          const hasPreviousFragment = index > 0 && address[line[index - 1].name]
```
lucasecdb, line 17: I'm thinking about refactoring this function, `PlaceDetails` and `AddressForm` later because this "rendering" logic of the address is the same among all of them, they just differ on what is the output (for this function it's a string, for place details it's spans and for address form it's inputs)

react/formatAddressToString.ts
```diff
@@ -0,0 +1,43 @@
+import { Address } from 'vtex.checkout-graphql'
+import { CountryRules } from 'vtex.address-context/types'
+import msk from 'msk'
+
+export default function formatAddressToString(
+  address: Address,
+  rules: CountryRules
+) {
+  const displaySpec = rules.display.compact
```
wdsrocha, line 9: how about receiving a display type as a prop? In the future we might want to get the string for a `minimal` display, for example

lucasecdb: nice! will do

lucasecdb: done in f92cd9f

wdsrocha (approved):
boa

lucasecdb:
@jeffersontuc @alvimm can you review?

jeffersontuc (approved):

PR #35, opened 2020-12-08
CHK-435: Session token

wdsrocha:

Many geolocalization providers allows the use of a session token in order to keep track of a single "autocomplete usage session". This usually has the benefits of being billed only once, which can save us some costs. More info on the Jira task.

This PR handles the usage of session tokens, generating one token when starting to use the component and refreshing this token when the user selects an address.

- [Add something to the cart](https://geolocation--checkoutio.myvtex.com/cart/add?sku=289)
- Use a new address on cart's shipping
- Open devtools
	- Type an address slowly. A single token should be logged multiple times, one for each stroke that sends a query
	- Select a suggested address. The same token should be logged one last time
	- Go back to the NewAddressForm and type a new address. It should log a different token.

obs: there will be no token logs at production level.

 (tip of the day: don't host your gifs in external services 😢 )

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.


wdsrocha:
PR is really huge because it's rebased above another PR. Will merge the other one before finishing this :-(

lucasecdb:
@wdsrocha you can base this pr to the branch it depends on, and when you merge the other pr, github will automatically change the base branch of this pr to master

wdsrocha:
Cool, good to know!

react/LocationSearch.tsx
```diff
@@ -149,7 +164,8 @@ const LocationSearch: React.FC<LocationSearchProps> = ({
     }
 
     setSearchTerm(selectedAddress)
-    executeGetAddress({ variables: { id } })
+    executeGetAddress({ variables: { id, sessionToken } })
+    refetchSessionToken()
```
wdsrocha, line 168: Is there any drawback in using `useQuery` + `refetch` versus `useLazyQuery`? I'm just using it because I wanted the first query to be executed on startup.

react/LocationSearch.tsx
```diff
@@ -111,7 +115,18 @@ const LocationSearch: React.FC<LocationSearchProps> = ({
...
+    loading: loadingSessionToken,
+    refetch: refetchSessionToken,
+  } = useQuery<Query, {}>(SESSION_TOKEN)
+  const sessionToken = loadingSessionToken
+    ? null
+    : sessionTokenData?.sessionToken ?? null
+  const [suggestions, loadingSuggestions] = useSuggestions(
+    debouncedSearchTerm,
+    sessionToken
+  )
```
wdsrocha, line 129: Maybe I'm overthinking here, but this is starting to get bloated. Does anyone have tips or suggestions on how to organize this?

wdsrocha: I was going to create an `useAddressAutocomplete` hook, similar to [Google's useAutocomplete](https://material-ui.com/pt/components/autocomplete/#customized-hook) or [this dude's usePlacesAutocomplete](https://dev.to/wellyshen/introducing-use-places-autocomplete-react-hook-for-google-maps-places-autocomplete-41h5), and although I would be able to hide at least the `sessionToken` from the component, the bloat would just be moved inside a hook (I guess). What do you ppl think?

react/LocationSearch.tsx
```diff
@@ -111,20 +115,31 @@ const LocationSearch: React.FC<LocationSearchProps> = ({
     searchTerm,
     DEBOUNCE_DELAY_IN_MS
   )
-  const [suggestions, loadingSuggestions] = useSuggestions(debouncedSearchTerm)
+  const {
+    data: sessionTokenData,
+    loading: loadingSessionToken,
+    refetch: refetchSessionToken,
+  } = useQuery<Query, {}>(SESSION_TOKEN)
```
lucasecdb, line 122: you should probably use the `notifyOnNetworkStatusChange` option here if you want the `loadingSessionToken` to be `true` after you call `refetchSessionToken`

react/__tests__/LocationSelect.test.tsx
```diff
@@ -7,9 +7,10 @@ import LocationSelect from '../LocationSelect'
 import {
   sampleAddress as addressWithoutLocationSelect,
   incompleteAddress,
-} from '../__mocks__/mockDescriptions'
+} from '../__fixtures__/address.fixture'
 
-describe('Location Select', () => {
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('Location Select', () => {
```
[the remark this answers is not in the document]
wdsrocha: damn :/

react/__tests__/LocationSearch.test.tsx
```diff
@@ -23,6 +25,17 @@ interface RenderComponentProps {
...
 
+const defaultGraphqlMocks = [
+  googleLogo,
+  defaultSessionToken,
+  // This duplication is intentional: the first one will be used during the
+  // first query and the second one is used during the refetch. Currently,
+  // Apollo provides not many good alternatives to do that. You can read
+  // more about it on this issue:
+  //  https://github.com/apollographql/apollo-client/issues/7286
+  defaultSessionToken,
```
wdsrocha, line 36: That looks bad. Also, after setting [`notifyOnNetworkStatusChange` to true](https://github.com/vtex-apps/place-components/pull/35/files#diff-735283229f349f8d66e9228f68d569dfae0d9f094d13b482e00b694e9f12f7aeR123) I had to create a [`simpleSuggestionWithoutToken`](https://github.com/vtex-apps/place-components/pull/35/files#diff-3ad87d3171f462fcfe0d8afa2cf19ac4ebdc29cef0ce6f5310de0db95b81b857R46-R55) because the second query returns `null` (which doesn't makes sense in my opinion).

Any clue on how to improve this? cc: @lucasecdb

wdsrocha: Besides that, I've tried all the [implementation suggestions here](https://stackoverflow.com/questions/58981274/is-there-a-way-to-mock-refetch-in-mockedprovider-apollo-client) (this stack overflow link is referenced on the issue linked in the comment btw) and none of them worked (might be my fault though).

wdsrocha:
Just rebased the commits in order to make it easier to review. No "real" change added after all those force pushes

wdsrocha:
Hold on a second, my workspace blew up and I'm setting up another one

wdsrocha:
The whole point of bc1181d is to prevent new queries being called as soon as an address is selected. Before this commit, the flow was:

1. Use selects an address
2. The input text (`searchTerm`) change it's value to the one that was selected
3. That fires two useless queries: one for the text change and the other for the `sessionToken` refresh

My solution for this is to have a `displayedSearchTerm`, that is the real value that will be displayed for the user, but not necessarily will change `searchTerm` and trigger the query. For the `sessionToken` part, I think it can just be removed from the effect hook dependency.

jeffersontuc (approved):
LGTM

lucasecdb:
@wdsrocha I have one question, why do we needed to create a query for the session token if we could _theoretically_ genereate it on the frontend?

wdsrocha:
@lucasecdb because different providers might require specific session tokens or no tokens at all. Currently, the only provider that seems to require a session token is Google Maps, so it might not even be a good reason after all :/

lucasecdb (approved):

PR #37, opened 2020-12-28
CHK-469: Address suggestions restricted by country

wdsrocha:

LocationSearch passes country from address context to addressSuggestions query, restricting the results to the specified country. Depends on https://github.com/vtex-apps/geolocation-graphql-interface/pull/7

- [Add global product to cart](https://geolocation--checkoutio.myvtex.com/cart/add?sku=312)
- Choose delivery options
- Check how suggestions change based on the country you choose

Obs: it's safe to test with Brazil and USA, other countries might crash.

on Brazil | on USA
-|-
image | image

#### Notes

I don't think there are any relevant tests to add.

### DIFF
--- a/react/LocationCountry.tsx
+++ b/react/LocationCountry.tsx
@@ -116,7 +116,6 @@ const LocationCountry: React.FC<LocationCountryProps> = ({
       <ListboxInput
         translate={undefined}
         aria-labelledby={labelId}
-        className="h-100"
         value={country}
         onChange={setCountry}
       >
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `LocationSearch` results can now be restricted per country.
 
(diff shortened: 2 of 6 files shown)


react/LocationCountry.tsx
```diff
@@ -116,7 +116,6 @@ const LocationCountry: React.FC<LocationCountryProps> = ({
       <ListboxInput
         translate={undefined}
         aria-labelledby={labelId}
-        className="h-100"
```
wdsrocha, line 119: For some reason, I added this a while ago, but now it's breaking on cart shipping

image

After undoing this change, the component seems fine on cart and checkout 🤷🏾

lucasecdb: nice :satisfaction:

lucasecdb (approved):

react/LocationCountry.tsx
alvim: The delivery options in cart in Safari is broken:

image

I don't know if it's this PR's scope (since the whole shipping step in checkout is broken too), but it worth checking.

wdsrocha: I'll take a look

wdsrocha: It's the device coordinates component again :-/
We should definitely open a task for that, but this isn't related to this PR indeed

alvim (approved):

PR #38, opened 2020-12-30
Upgrade packages and fix linter issues

wdsrocha:

As title says. After upgrading the packages, **lots** of linter issues appeared, so this PR is an attempt to fix them.

[Workspace](https://geolocation--checkoutio.myvtex.com/cart/add?sku=312)


react/LocationSelect.tsx
```diff
@@ -112,7 +115,10 @@ const LocationSelect: React.FC = () => {
         </div>
       )
 
-      currentOptions = fieldValue ? currentOptions[fieldValue as keyof {}] : {}
+      currentOptions = fieldValue
+        ? // eslint-disable-next-line @typescript-eslint/ban-types
```
lucasecdb, line 119: I hate this rule 😆

lucasecdb (approved):

react/LocationSelect.tsx
wdsrocha: tbh I don't really get the difference of `object` vs `Record<string, unknown>` (not that I actually really searched for it tho), but what scared me a little bit is that `currentOptions` can also be an array, as shown on line 90

jeffersontuc (approved):

PR #39, opened 2020-12-30
CHK-488: Fix geolocation on Safari

wdsrocha:

Shipping breaks on safari due to a bug in device coordinate component.

image

Turns out that the problem is because some browsers don't support asking for `permission`. See more about that [here](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/permissions#Browser_compatibility).

- open safari on pc
- [workspace](https://geolocation--checkoutio.myvtex.com/cart/add?sku=312)
- click on delivery options and click on device coordinate

It should prompt for your geolocation and work normally.

#### Notes

Should we create a unit test for this? I feel like an e2e test for safari with complete purchase flow would be enough, but idk.

### DIFF
--- a/react/DeviceCoordinates.tsx
+++ b/react/DeviceCoordinates.tsx
@@ -84,6 +84,11 @@ const DeviceCoordinates: React.FC<Props> = ({ onSuccess }) => {
   }
 
   useEffect(() => {
+    // some browsers do not have/require permissions
+    if (!navigator.permissions) {
+      return
+    }
+
     navigator.permissions
       .query({ name: 'geolocation' })
       .then((result: PermissionStatus) => {
--- a/react/__tests__/DeviceCoordinates.test.tsx
+++ b/react/__tests__/DeviceCoordinates.test.tsx
@@ -1,0 +1,164 @@
+import React from 'react'
+import { fireEvent, render, screen, waitFor } from '@vtex/test-tools/react'
+import { MockedProvider, MockedResponse } from '@apollo/react-testing'
+import { Address } from 'vtex.checkout-graphql'
+import { AddressRules } from 'vtex.address-context/types'
+
+import { AddressContextProvider } from '../__mocks__/vtex.address-context/AddressContext'
+import DeviceCoordinates from '../DeviceCoordinates'
+import { reverseGeocode } from '../__fixtures__/graphql/reverseGeocode.fixture'
+import { vtexOfficeAddress } from '../__fixtures__/address.fixture'
+
+interface RenderComponentProps {
+  deviceCoordinatesProps?: React.ComponentProps<typeof DeviceCoordinates>
+  address?: Address | null
+  countries?: string[]
+  rules?: AddressRules
+  graphqlMocks?: MockedResponse[]
+}
+
+const defaultGraphqlMocks = [] as MockedResponse[]
+
+describe('Device Coordinates', () => {
+  const renderComponent = ({
+    deviceCoordinatesProps,
+    address = { country: 'BRA' },
+    countries = [],
+    rules = {},
+    graphqlMocks = [],
+  }: RenderComponentProps) => {
+    return render(
+      <MockedProvider
+        addTypename={false}
+        mocks={[...defaultGraphqlMocks, ...graphqlMocks]}
+      >
+        <AddressContextProvider
+          address={address}
+          countries={countries}
+          rules={rules}
+        >
+          <DeviceCoordinates {...deviceCoordinatesProps} />
+        </AddressContextProvider>
+      </MockedProvider>
+    )
+  }
+
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+
+  describe('when geolocation is granted', () => {
+    it('should get an address from coordinates', async () => {
+      const callOrder: string[] = []
+
+      const { query } = navigator.permissions
+      const mockedPermissionsQuery = query as jest.MockedFunction<typeof query>
+
+      mockedPermissionsQuery.mockImplementation(_permission => {
+        callOrder.push('permissions')
+
+        return Promise.resolve({ state: 'granted' } as PermissionStatus)
+      })
+
+      Object.defineProperty(navigator, 'permissions', {
+        get() {
+          return { query: mockedPermissionsQuery }
+        },
+      })
+
+      const { getCurrentPosition } = navigator.geolocation
+      const mockedGetCurrentPosition = getCurrentPosition as jest.MockedFunction<
+        typeof getCurrentPosition
+      >
+
+      mockedGetCurrentPosition.mockImplementation(success => {
+        callOrder.push('geolocation')
+
+        return Promise.resolve(
+          success({
+            coords: {
+              latitude: vtexOfficeAddress.geoCoordinates[0],
+              longitude: vtexOfficeAddress.geoCoordinates[1],
+            },
+          } as GeolocationPosition)
+        )
+      })
+
+      Object.defineProperty(navigator, 'geolocation', {
+        get() {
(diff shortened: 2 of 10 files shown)


jeffersontuc (approved):

react/DeviceCoordinates.tsx
```diff
@@ -84,13 +84,21 @@ const DeviceCoordinates: React.FC<Props> = ({ onSuccess }) => {
...
+        .then((result: PermissionStatus) => {
+          if (result.state === 'denied') {
+            setGeolocationPermission('denied')
+          }
+        })
+    } catch (e) {
+      // ignoring TypeError handles browsers that don't support `permission`
+      // https://developer.mozilla.org/en-US/docs/Web/API/Navigator/permissions#Browser_compatibility
+      if (!(e instanceof TypeError)) {
+        throw e
```
wdsrocha, line 99: is it ok to rethrow like that?

alvim: If with "rethrow like that" you mean like "catching only to throw again", this is not a problem in JS. Unlike other languages where the stack trace would be modified, in JS it will not. It would be just as good without the `try catch`.

But I think that it's not a good practice to implement flow control inside a catch block. It's made for exception handlings, not flow control. I think you should check if the `navigator` has `permissions` then execute the methods, not intentionally generate an error (in some cases).

Let me know if I'm being clear and if you agree.

wdsrocha: >  [...] It would be just as good without the try catch.

Cool!

> But I think that it's not a good practice to use flow control inside a catch block [...]

Agreed, I just don't know exactly how to test for that...

wdsrocha: ah, nvm, I can just check if it's undefined. I was thinking that I would need to know which browser the user is using, but that does not seem to be the case

lucasecdb (approved):

alvim (dismissed):

react/__tests__/DeviceCoordinates.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <AddressContextProvider
+          address={address}
+          countries={countries}
+          rules={rules}
+        >
+          <DeviceCoordinates {...deviceCoordinatesProps} />
+        </AddressContextProvider>
+      </MockedProvider>
+    )
+  }
```
wdsrocha, line 44: Although this is very similar to LocationSearch renderComponent, I think there is no need to merge this in an util class, at least for now.

react/__tests__/DeviceCoordinates.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+        expect(mockedOnSuccess).toHaveBeenCalledTimes(1)
+      })
+    })
+  })
+
+  // TODO
+  describe('when geolocation is blocked', () => {})
+
+  // TODO
+  describe('when geolocation is prompt', () => {})
```
wdsrocha, line 159: Those are all possible remaining states for geolocalization so I think it would be good to test them in the future

react/setupTests.js
```diff
@@ -0,0 +1,7 @@
+global.navigator.geolocation = {
+  getCurrentPosition: jest.fn(),
+}
+
+global.navigator.permissions = {
+  query: jest.fn(),
+}
```
wdsrocha, line 7: I wish it could work like this:

```suggestion
global.navigator = {
  geolocation: {
  	getCurrentPosition: jest.fn(),
  },
  permissions: {
	query: jest.fn(),
  }
}
```

but it simply makes `geolocation` and `permissions` undefined 🤷🏾 
no big deal, but if anyone knows why this doesn't works, I would like to know

jeffersontuc: Did you try something like this

```
global.navigator = () => {
  const geolocation = {
    getCurrentPosition: jest.fn(),
  }

  const permissions = {
    query: jest.fn(),
  }

  return {
    geolocation,
    permissions
  }
}
```

wdsrocha: no but hmm, let me try

wdsrocha: didn't worked, same problem:

`TypeError: Cannot read property 'query' of undefined` (same for `getCurrentPosition`)

I think it kind of makes sense, because navigator isn't a function

jeffersontuc (approved):

react/setupTests.js
jeffersontuc: And like that? 🤔 

```
global.navigator = (() => {
  const geolocation = {
    getCurrentPosition: jest.fn(),
  }

  const permissions = {
    query: jest.fn(),
  }

  return {
    geolocation,
    permissions
  }
}())
```

jeffersontuc: But, personally, I don't see a problem in keeping it the way you did.

wdsrocha: I was really thinking that this one would work, but nope. Same error :thonking:

react/__tests__/DeviceCoordinates.test.tsx
alvim: Nice!

alvim (approved):

react/__tests__/DeviceCoordinates.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  afterEach(() => {
```
lucasecdb, line 161: can you move this up to the beginning of the function?

react/setupTests.js
lucasecdb: shouldn't this file be `setupTestsAfterEnv.js`?

react/__tests__/DeviceCoordinates.test.tsx
wdsrocha: Sure. Do you know if there is any eslint rule for that kind of thing?

wdsrocha:
> shouldn't this file be setupTestsAfterEnv.js?

not necessarily, because I believe this doesn't matter for this simple setting. Doc uses a generic name for the "after env setup" file as well: https://jestjs.io/docs/en/configuration#setupfilesafterenv-array

but still, do you prefer me to change? I have no strong opinion about this

lucasecdb:
> but still, do you prefer me to change? I have no strong opinion about this

I thought that maybe changing this would fix the issues you posted 😄 but if it doesn't, I don't mind keeping it as is

react/__tests__/DeviceCoordinates.test.tsx
lucasecdb: I don't think so :-/

react/package.json
```diff
@@ -33,5 +33,10 @@
     "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.126.5/public/@types/vtex.render-runtime",
     "vtex.styleguide": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.styleguide@9.134.1/public/@types/vtex.styleguide"
   },
+  "jest": {
+    "setupFilesAfterEnv": [
+      "<rootDir>/setupTests.js"
+    ]
+  },
```
wdsrocha, line 40: > I thought that maybe changing this would fix the issues you posted 😄 but if it doesn't, I don't mind keeping it as is

@lucasecdb, this would be equivalent to renaming, no? But now I get your point

lucasecdb: ohh, it already is a after env setup file, right? yeah, the rename wouldn't matter then 😂

wdsrocha: I tested it anyway and it does not solve :(
but ty for the suggestion

wdsrocha: (btw sent last message before seeing your reply :slowpoke:)

lucasecdb (approved):

PR #43, opened 2021-02-03
Fix empty fields hidden in address form component

lucasecdb:

Fixes an issue where empty fields with `null` were being hidden instead of shown as an empty input.

[Workspace](https://addr--checkoutio.myvtex.com/cart/add?sku=289).

You can see that in the address step of checkout the complement will always be shown.

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.
- [ ] Linked this PR to a Jira story (if applicable).

#### Type of changes

✔️ | Type of Change
---|---
✔️ | Bug fix 
_ | New feature 
_ | Breaking change 
_ | Technical improvements

### DIFF
--- a/react/AddressForm.tsx
+++ b/react/AddressForm.tsx
@@ -190,8 +190,7 @@ const AddressForm: React.FC<AddressFormProps> = ({
               if (
                 !field ||
                 ignoredFields.has(fragment.name) ||
-                hiddenFields.includes(fragment.name) ||
-                address[fragment.name] == null
+                hiddenFields.includes(fragment.name)
               ) {
                 return false
               }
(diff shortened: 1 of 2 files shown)


alvim (approved):

jeffersontuc (approved):

react/AddressForm.tsx
```diff
@@ -219,7 +218,7 @@ const AddressForm: React.FC<AddressFormProps> = ({
                   }))
                 }
 
-                const value = address[fragment.name] as string
+                const value = address[fragment.name] ?? ''
```
wdsrocha, line 221: What happens if the `fragment.name` is `geoCoordinates` (array of float) or `isDisposable` (boolean)?

lucasecdb: these fields shouldn't be editable, so we need to "curate" the countries-data repo to always mark them as `hidden: true`

lucasecdb: also, even before this PR you can see that the `value` variable was being "converted" to string with `as string`, so this shouldn' t change the behavior of the code compared to what is was before

PR #45, opened 2021-02-12
CHK-550: Properly handle error state on LocationInput

wdsrocha:

When inserting a postal code in the `LocationInput` component...

1. no error state appear, even with an wrongly formatted postal code
2. when submitting an invalid postal code for the second time in a row, the component loads indefinitely

[Workspace](https://postalcode--checkoutio.myvtex.com/cart/add?sku=312&sc=2)

First scenario (see the **screenshot** bellow):
1. select Brazil in the country combobox
2. insert the "22250-04" postal code and submit (obs: it's missing the last digit)
3. it should display an error message _immediately_

Second scenario (see the **video** bellow):
1. select USA in the country combobox
2. insert the "asdf" postal code and submit
3. it should display an error message
4. delete the last character, re-add it and submit
5. it should display an error message, instead of loading indefinitely

Obs: the reason why you should test the second scenario with USA is because this country still don't have a validation pattern for the front-end, so the component does not block the submission button right away - it consults the back-end instead.

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

image

https://user-images.githubusercontent.com/26108090/108114818-a82af380-706f-11eb-96ab-7cc8396c7679.mov

### DIFF
--- a/react/LocationInput.tsx
+++ b/react/LocationInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useCallback, useRef } from 'react'
 import { useAddressContext } from 'vtex.address-context/AddressContext'
 import { IconSearch, Input, Button, ButtonPlain } from 'vtex.styleguide'
 import { FormattedMessage, useIntl } from 'react-intl'
@@ -27,10 +27,13 @@ const LocationInput: React.FC<Props> = ({
   const { culture } = useRuntime()
   const intl = useIntl()
   const [inputValue, setInputValue] = useState('')
+  // the network-only fetch policy asserts that if an incorrect postal code
+  // has been submitted twice in a row, it won't return the cached error,
+  // re-triggering the effect hook and avoiding an infinite loop state
   const [executeGetAddressFromPostalCode, { error, data }] = useLazyQuery<
     Query,
     QueryGetAddressFromPostalCodeArgs
-  >(GET_ADDRESS_FROM_POSTAL_CODE)
+  >(GET_ADDRESS_FROM_POSTAL_CODE, { fetchPolicy: 'network-only' })
 
   const [loading, setLoading] = useState(false)
   const [invalidPostalCode, setInvalidPostalCode] = useState(false)
@@ -38,6 +41,34 @@ const LocationInput: React.FC<Props> = ({
   const country = address?.country ?? culture.country
 
   const countryRules = country ? rules[country] : undefined
+
+  const prevCountryRulesRef = useRef(countryRules)
+
+  const formatAndValidate = useCallback(() => {
+    const { mask, pattern } = countryRules?.fields.postalCode ?? {}
+
+    if (!mask) {
+      return
+    }
+
+    setInputValue(msk(inputValue, mask))
+    const patternInvalid =
+      inputValue.length && pattern && !inputValue.match(pattern)
+
+    if (patternInvalid) {
+      setInvalidPostalCode(true)
+    }
+  }, [countryRules?.fields.postalCode, inputValue])
+
+  useEffect(() => {
+    if (prevCountryRulesRef.current === countryRules) {
+      return
+    }
+
+    prevCountryRulesRef.current = countryRules
+    setInvalidPostalCode(false)
+    formatAndValidate()
+  }, [countryRules, formatAndValidate])
 
   useEffect(() => {
     let cancelled = false
@@ -68,7 +99,7 @@ const LocationInput: React.FC<Props> = ({
     }
 
     if (error) {
-      console.warn(error.message)
+      console.error(error.message)
       setInvalidPostalCode(true)
       setLoading(false)
     }
@@ -80,10 +111,13 @@ const LocationInput: React.FC<Props> = ({
 
   const handleSubmit: React.EventHandler<
     React.FormEvent<HTMLFormElement> | React.MouseEvent<HTMLButtonElement>
-  > = evt => {
-    evt.preventDefault()
+  > = event => {
+    event.preventDefault()
+    if (invalidPostalCode) {
+      return
+    }
+
     setLoading(true)
-    setInvalidPostalCode(false)
     executeGetAddressFromPostalCode({
       variables: {
         postalCode: inputValue,
@@ -92,18 +126,18 @@ const LocationInput: React.FC<Props> = ({
     })
   }
 
-  const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = evt => {
-    setInputValue(evt.target.value)
-  }
-
-  const handleInputBlur = () => {
-    const postalCodeMask = countryRules?.fields.postalCode?.mask
-
-    if (!postalCodeMask) {
-      return
+  const handleInputChange: React.ChangeEventHandler<HTMLInputElement> = event => {
+    if (invalidPostalCode) {
+      setInvalidPostalCode(false)
     }
 
-    setInputValue(msk(inputValue, postalCodeMask))
+    setInputValue(event.target.value)
+  }
+
+  const handleKeyDown: React.KeyboardEventHandler<HTMLInputElement> = event => {
+    if (event.key === 'Enter') {
+      formatAndValidate()
+    }
   }
 
   if (!countryRules) {
@@ -130,7 +164,8 @@ const LocationInput: React.FC<Props> = ({
           size="large"
           value={inputValue}
           onChange={handleInputChange}
-          onBlur={handleInputBlur}
+          onBlur={formatAndValidate}
+          onKeyDown={handleKeyDown}
           disabled={loading}
           error={invalidPostalCode}
           errorMessage={
--- a/react/__tests__/LocationInput.test.tsx
+++ b/react/__tests__/LocationInput.test.tsx
@@ -1,0 +1,136 @@
+import React from 'react'
+import { render, screen, waitFor } from '@vtex/test-tools/react'
+import { MockedResponse } from '@apollo/react-testing'
+import { Address } from 'vtex.checkout-graphql'
+import { AddressRules, CountryRules, Field } from 'vtex.address-context/types'
+import { AddressContextProvider } from 'vtex.address-context/AddressContext'
+import userEvent from '@testing-library/user-event'
+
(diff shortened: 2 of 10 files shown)


react/LocationInput.tsx
```diff
@@ -93,17 +100,29 @@ const LocationInput: React.FC<Props> = ({
...
+      setInvalidPostalCode(false)
+    }
+
     setInputValue(evt.target.value)
   }
 
-  const handleInputBlur = () => {
+  const isPatternInvalid = () =>
+    countryRules?.fields.postalCode?.pattern != null &&
+    inputValue.match(countryRules.fields.postalCode.pattern) === null
```
wdsrocha, line 112: Just realized that this could be a const inside `formatAndValidate`, but that don't seems to matter too much

react/LocationInput.tsx
```diff
@@ -130,7 +149,8 @@ const LocationInput: React.FC<Props> = ({
           size="large"
           value={inputValue}
           onChange={handleInputChange}
-          onBlur={handleInputBlur}
+          onBlur={formatAndValidate}
+          onKeyDown={formatAndValidate}
```
wdsrocha, line 153: Added this for a11y matters (hitting the enter key). I suppose those `onBlur` and `onKeyDown` cover all scenarios where we need to format and validate before submitting

lucassuzano:
Bug:
1. select USA in the country combobox
2. insert any postal code and submit
3. choose to change the postal code
4. Show the same country so remaining USA select
_Here the system change to Brazil_ 

1. select USA in the country combobox
2. insert any postal code 
3. change to another country
4. System change the country and remaining with the same postal code
5. System validate the postal code for the new country
_The system don't do this last step, they don't validate the postal code with the new country_

Importante validar os fluxos não só com EUA e Brazil, para entender se as incosistências são particulares desses países

Deixar claro qual a experiência de acordo com REGEX
 
Países sem REGEX terão o fluxo XPTO
Países com REGEX terão fluxo XPTO

wdsrocha:
> Scenario 1

Not really related to this PR. We will create a task on jira to handle this later on

> Scenario 2

Fixed. Could you test it again, please?

> Importante validar os fluxos não só com EUA e Brazil, para entender se as incosistências são particulares desses países

I don't think it really matters, what is most important for the LocationInput scenario is if the country settings have input mask or not. I'm using Brazil and USA just because Brazil have mask and USA don't

> Países sem REGEX terão o fluxo XPTO

Só serão validados no backend (nada de erro durante o blur do componente)

> Países com REGEX terão fluxo XPTO

Terá validação no frontend também

Não sei como deixar mais explícito esse fluxo. Tem ideia? Ou nem precisa?

react/LocationInput.tsx
```diff
@@ -39,14 +42,39 @@ const LocationInput: React.FC<Props> = ({
...
+    if (patternInvalid) {
+      setInvalidPostalCode(true)
+    }
+  }
+
+  useEffect(() => {
+    setInvalidPostalCode(false)
+    formatAndValidate()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [countryRules])
```
lucasecdb, line 68: why disabling the exhaustive deps here? if you remove this comment and use `useCallback` in the `formatAndValidate` function you would get the same behavior (because the function actually depends on `countryRules`)

react/LocationInput.tsx
```diff
@@ -39,14 +42,39 @@ const LocationInput: React.FC<Props> = ({
...
+
   useEffect(() => {
     let cancelled = false
 
     if (data?.getAddressFromPostalCode) {
       // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
       ;(onSuccess?.(data.getAddressFromPostalCode) || Promise.resolve())
         .then(() => {
-          setAddress(prevAddress => ({
+          setAddress((prevAddress) => ({
```
lucasecdb, line 77: did you update the lint packages? it seems it isn't updated in the GH actions, maybe some file is being ignored by gitignore?

react/__tests__/LocationInput.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+    renderComponent({
+      locationInputProps: { onSuccess: mockedOnSuccess },
+      graphqlMocks: [validPostalCode],
+    })
+
+    const input = screen.getByRole('textbox', {
+      name: /postal code/i,
+    })
+
+    input.focus()
```
lucasecdb, line 78: why aren't you using `fireEvent` here? using the focus method on the dom element directly would require us to wrap this in `act` else we would get that act warning in the test run

this also applies to the other uses of the `focus` method in this test file

react/__tests__/LocationInput.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+    fireEvent.change(input, { target: { value: '22250040' } })
+
+    const button = screen.getByRole('button')
+
+    // formatting happens on input blur, so this should be simulated manually
+    // see https://testing-library.com/docs/guide-events/#interactions-vs-events
+    button.focus()
```
lucasecdb, line 85: same as above, but you could explicitly blur the input here instead of focusing on the button, or use the [`user-event`](https://github.com/testing-library/user-event) library

lucasecdb (commented):
can you also create some tests to assert that the mask is only being applied on the blur event?

react/LocationInput.tsx
wdsrocha: I've updated in another branch, weird. The GH actions should run the same linter as the current repo state, right? instead of the newest VTEX linter version

react/LocationInput.tsx
wdsrocha: the thing is, with `useCallback` and adding `formatAndValidate` to the deps of `useEffect` results in the effect being triggered everytime `formatAndValidate` gets called. is there any other alternative?

react/__tests__/LocationInput.test.tsx
wdsrocha: weird, I didn't got act warnings, so I thought that it was ok. The reason why I'm not using `fireEvent` directly was due to the [link I've linked bellow](https://testing-library.com/docs/guide-events/#interactions-vs-events), but tbh I think this line in specific seems unecessary. Only the one in line 85 makes sense I guess

react/__tests__/LocationInput.test.tsx
wdsrocha: this lib looks pretty good. I'll try it!

react/LocationInput.tsx
lucasecdb: I see, that's because `formatAndValidate` depends on `inputValue`, right? you could update it to instead of using the value from the closure to use it from an argument, then you wouldn't have this infinite loop

wdsrocha:
> can you also create some tests to assert that the mask is only being applied on the blur event?

plus on the `onKeyDown` event, right?

react/LocationInput.tsx
wdsrocha: ohh that's right. I'll change it

wdsrocha: but wait, now the useEffect depends on `inputValue` in order to send it to `formatAndValidate`. Am I missing something?

wdsrocha: here is my attempt

```typescript
  const formatAndValidate = useCallback(
    (postalCode: string) => {
      // other stuff here
    },
    [
      countryRules?.fields.postalCode?.mask,
      countryRules?.fields.postalCode?.pattern,
    ]
  )

  useEffect(() => {
    setInvalidPostalCode(false)
    formatAndValidate(inputValue) // <- how I should handle this?
  }, [countryRules, formatAndValidate])
```

lucasecdb: you can use a ref to only trigger the effect when the `countryRules` dependency changes, e.g:

```js
const formatAndValidate = useCallback(/* ... */)

const prevCountryRulesRef = useRef(
  null // Use null if you want the effect to trigger on first render, else use `countryRules`
)

useEffect(() => {
  if (prevCountryRulesRef.current === countryRules) {
    return
  }

  prevCountryRulesRef.current = countryRules

  setInvalidPostalCode(false)
  formatAndValidate(inputValue)
}, [
/* dependencies doesn't matter anymore, since you are
 * now only running the effect when there is a change to
 * the countryRules ref, so you can safely include any deps
 * the eslint rule wants to
 */
])
```

wdsrocha:
Done, @lucasecdb, check if it's ok, please. Btw, I'll squash some commits if accepted.

react/__tests__/LocationInput.test.tsx
```diff
@@ -0,0 +1,136 @@
...
+    expect(input).toHaveValue('2225004')
+
+    userEvent.click(screen.getByRole('button'))
+    expect(input).toHaveValue('22250-04')
+
+    userEvent.type(input, '{selectall}2225004')
+    expect(input).toHaveValue('2225004')
+
+    userEvent.type(input, '{enter}')
+    expect(input).toHaveValue('22250-04')
```
lucasecdb, line 134: :satisfaction:

lucasecdb (approved):
LGTM! well done ✨

react/LocationInput.tsx
```diff
@@ -80,8 +111,12 @@ const LocationInput: React.FC<Props> = ({
...
-  > = evt => {
-    evt.preventDefault()
+  > = event => {
+    event.preventDefault()
+    if (invalidPostalCode) {
+      return
+    }
+
     setLoading(true)
     setInvalidPostalCode(false)
```
jeffersontuc, line 121: I think this line isn't necessary anymore, right?

jeffersontuc: Since if `invalidPostalCode` is `true`, it will return early. Or am I missing something?

jeffersontuc (approved):

react/__tests__/LocationInput.test.tsx
wdsrocha: also, loved this lib :amaze:

react/LocationInput.tsx
wdsrocha: Nice catch! Will remove

PR #46, opened 2021-02-25
Update linter, typings and refactor country usage

wdsrocha:

Just some minor things to improve `place-components` code quality.

You can test the updated components in [this workspace](https://postalcode--checkoutio.myvtex.com/cart). It might be tricky to find bugs that way, so I suggest to focus more in the code itself.


react/PlaceDetails.tsx
```diff
@@ -40,31 +43,37 @@ const PlaceDetails: React.FC<Props> = ({
...
+              !hiddenFields.includes(fragment.name as FieldName)
           )
         })
         .filter((line) => line.length > 0)
         .map((line, displayIndex) => (
           <div className="flex" key={displayIndex}>
             {line.map((fragment, index) => {
+              if (typeof index !== 'number') {
+                return
+              }
```
wdsrocha, line 58: I don't know why this is needed... index is clearly a `number` already, but eslint complains with [this](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/restrict-plus-operands.md) anyway.

react/formatAddressToString.ts
```diff
@@ -1,37 +1,50 @@
...
+      line.filter(({ name }) => isAddressField(name) && address[name])
+    )
     .filter((line) => line.length > 0)
     .map((line) =>
       line
         .map((fragment, index) => {
-          const hasPreviousFragment = index > 0 && address[line[index - 1].name]
+          if (typeof index !== 'number') {
+            return
+          }
```
wdsrocha, line 28: :ditto:

react/formatAddressToString.ts
```diff
@@ -1,37 +1,50 @@
...
+      line.filter(({ name }) => isAddressField(name) && address[name])
+    )
+    .filter((line) => line.length > 0)
+    .map((line) =>
       line
         .map((fragment, index) => {
-          const hasPreviousFragment = index > 0 && address[line[index - 1].name]
+          if (typeof index !== 'number') {
+            return
+          }
```
lucasecdb, line 28: how is this not a number? 🤔

react/useCountry.ts
```diff
@@ -0,0 +1,13 @@
...
+import { useAddressContext } from 'vtex.address-context/AddressContext'
+
+export const useCountry = () => {
+  const { culture } = useRuntime()
+  const { address } = useAddressContext()
+
+  // TODO: should also try to get country from:
+  // - addresses in account
+  // - IP
+  // - store config
```
wdsrocha, line 11: This comments already existed and I just bring them here. Does anyone knows if we have jira tasks for those already?

react/formatAddressToString.ts
wdsrocha: ikr lol: https://github.com/vtex-apps/place-components/pull/46#discussion_r582963879

lucasecdb: that's weird, maybe something is messed up in our tsconfig..

react/useCountry.ts
lucasecdb: we do not, also I think "store config" is the one we get from `culture.country` already

react/formatAddressToString.ts
lucasecdb: can you add a comment above this if so we know we should remove it in the future (if we find a solution)?

wdsrocha: sure

wdsrocha: for some reason it's fixed ¯\_(ツ)_/¯
I'm clueless why that was happening but I'll just accept it

jeffersontuc (approved):
Nice job!

react/formatAddressToString.ts
lucasecdb: oh, better yet then 😄

lucasecdb (approved):

PR #47, opened 2021-03-04
CHK-568: Add optional custom address prop to place details

wdsrocha:

Sometimes we might want to render the details of places that aren't from the user, like for pickup store. Now it's also possible to use place details inside address list, if we want to.

Same as https://github.com/vtex-apps/checkout-shipping/pull/31

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,7 +22,7 @@ Country data is sorted according to [Figma designs](https://www.figma.com/file/u
 
 ## Place Details
 
-The Place Details component displays a summary of a place according to the address it receives from the wrapper orchestrator, and the description of the structure that receives as prop. There are three variations of PlaceDetails according to the amount of information that needs to be shown in a particular scenario. Below you can see a description of them.
+The Place Details component displays a summary of a place according to the address from the wrapper orchestrator or a custom address passed via props. There are three variations of PlaceDetails according to the amount of information that needs to be shown in a particular scenario. Below you can see a description of them.
 
 - **Extended**: complete address, with all information that's necessary for carriers and users to identify it.
 - **Compact**: displays details gathered through LocationForm or from its info.
--- a/react/PlaceDetails.tsx
+++ b/react/PlaceDetails.tsx
@@ -5,6 +5,7 @@ import { AddressFields } from 'vtex.address-context/types'
 import { ButtonPlain } from 'vtex.styleguide'
 import { useIntl, defineMessages } from 'react-intl'
 import { DisplayData } from 'vtex.country-data-settings'
+import { Address } from 'vtex.checkout-graphql'
 
 import { FieldName } from './useAddressForm'
 import { useCountry } from './useCountry'
@@ -13,6 +14,7 @@ interface Props {
   display?: keyof Omit<DisplayData, '__typename'>
   hiddenFields?: AddressFields[]
   onEdit?: () => void
+  customAddress?: Address
 }
 
 const messages = defineMessages({
@@ -26,8 +28,10 @@ const PlaceDetails: React.FC<Props> = ({
   display = 'extended',
   hiddenFields = [],
   onEdit,
+  customAddress,
 }) => {
-  const { address, rules } = useAddressContext()
+  const { address: contextAddress, rules } = useAddressContext()
+  const address = customAddress ?? contextAddress
   const country = useCountry()
   const countryRules = rules[country]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `PlaceDetails` allows the usage of a custom address.
 
 ## [0.14.2] - 2021-02-26
 ### Changed
-- All components request country information in a standardiezed way.
+- All components request country information in a standardized way.
 - Typings were updated to match original types.
 
 ## [0.14.1] - 2021-02-23


jeffersontuc (approved):

silviasfon (approved):

react/PlaceDetails.tsx
```diff
@@ -26,8 +28,10 @@ const PlaceDetails: React.FC<Props> = ({
   display = 'extended',
   hiddenFields = [],
   onEdit,
+  customAddress,
 }) => {
-  const { address, rules } = useAddressContext()
+  const { address: contextAddress, rules } = useAddressContext()
```
RodrigoCMoraes, line 33: [@question](https://conventionalcomments.org): what is this `rules`?

docs/README.md
```diff
@@ -22,7 +22,7 @@ Country data is sorted according to [Figma designs](https://www.figma.com/file/u
 
 ## Place Details
 
-The Place Details component displays a summary of a place according to the address it receives from the wrapper orchestrator, and the description of the structure that receives as prop. There are three variations of PlaceDetails according to the amount of information that needs to be shown in a particular scenario. Below you can see a description of them.
+The Place Details component displays a summary of a place according to the address from the wrapper orchestrator or a custom address passed via props. There are three variations of PlaceDetails according to the amount of information that needs to be shown in a particular scenario. Below you can see a description of them.
```
RodrigoCMoraes, line 25: [@praise](https://conventionalcomments.org): good to see docs being updated!

RodrigoCMoraes:
[@question](https://conventionalcomments.org): why we don't have some sort of tests (Unit Test, Integration Test, etc...) in this PR?

react/PlaceDetails.tsx
wdsrocha: `rules` are a set of address rules and settings specific to a country ([AddressRules typing](https://github.com/vtex-apps/address-context/blob/61cbaf1ed0a7cdbba942f8cd36a46647b12cb50c/react/types.tsx#L71-L73)). For example, you can check the brazilian rules [here](https://github.com/vtex-apps/countries-data/blob/main/packages/bra/vtex.country-data-settings/configuration.json). This file includes things used for input validation (like CPF, phone, etc) and which components we should render on screen. To better visualize this, the property [display](https://github.com/vtex-apps/countries-data/blob/main/packages/bra/vtex.country-data-settings/configuration.json#L167) on extended mode describes the format that we want to show, like in the pictures bellow. For UK users, it makes no sense to display number, as it is already handled by address line 1, so we hide this from the user.

All this information comes from the `useAddressContext`

BR | UK
-|-
image | image

Obs: first picture is outdated with the current setting (there is no `receiverName` property) and UK rules haven't been created yet.

docs/README.md
wdsrocha: thanks :^D

btw, [this bot](https://github.com/vtex-apps/place-components/pull/47#issuecomment-790680631) helps a lot to keep the documentation updated by forcing the PR creator to explicitly define the status of the docs

[See more about it](https://github.com/vtex-apps/docs-bot)

react/PlaceDetails.tsx
RodrigoCMoraes: Thanks for this great explanation!

wdsrocha:
> [@question](https://conventionalcomments.org): why we don't have some sort of tests (Unit Test, Integration Test, etc...) in this PR?

I thought it wasn't necessary, but thinking about it now, it would be worth it and quick to do. But as suggested by my peer in the other repository, this feature might not be needed at all, so I'll just close this PR and remove the necessity to even create a test :)

docs/README.md
RodrigoCMoraes: Awesome! That's really good to know that we have this solution. But we have some solution for [pending docs](https://github.com/vtex-apps/io-documentation/issues?page=4&q=is%3Aissue+is%3Aopen)?

RodrigoCMoraes:
> > [@question](https://conventionalcomments.org): why we don't have some sort of tests (Unit Test, Integration Test, etc...) in this PR?
> 
> I thought it wasn't necessary, but thinking about it now, it would be worth it and quick to do. But as suggested by my peer in the other repository, this feature might not be needed at all, so I'll just close this PR and remove the necessity to even create a test :)

Do you know if we have an initiative to add [coverage](https://martinfowler.com/bliki/TestCoverage.html) or something like this on CI? 
IMHO, we have more wins than losses adopting a [testing culture](https://martinfowler.com/articles/testing-culture.html). Do you know if we already had discussed this topic in past?

react/PlaceDetails.tsx
mredigonda: Agree, awesome explanation 🙂

Maybe a good place to write down some words about what these `rules` mean could be the README of this project: https://github.com/vtex-apps/address-context

Super happy that you are taking good care of this project Wes! 😸

docs/README.md
wdsrocha: We don't. It only creates the issue and never notifies the codeowners again 😞. Maybe this bot could be improved to send a warning from time to time? Does that make sense to you?

wdsrocha:
@RodrigoCMoraes 

> Do you know if we have an initiative to add coverage or something like this on CI?

Not _yet_, but having code coverage is definitely a desired future. I just created a ["testing improvements" doc](https://www.notion.so/vtexhandbook/Testing-Quick-Wins-Improvements-99bcd1aacebe40f48ad744546770d30f) for the Checkout Squad, which I'll add more details in how to handle this problem there.

> IMHO, we have more wins than losses adopting a testing culture. Do you know if we already had discussed this topic in past?

Yes, the front-end chapter is well aware of this and we are improving, but it's still not a strong culture and we lack proper tooling (for example, no code coverage on CI!). I believe that by improving the tooling and embracing this culture on my own team, we'll be able to extend our solutions to the whole chapter and give a good example on why the testing culture is worth it.

wdsrocha:
Closing this since @lucasecdb [proposed a good solution that does not requires the necessity to change this repository](https://github.com/vtex-apps/checkout-shipping/pull/31#discussion_r595964795).

PR #50, opened 2022-01-31
Add Arabic and Norwegian variant translations

camilavcoutinho:
**What is the purpose of this pull request?**
Add Norwegian locale variant (nn) and Arabic translation.

**What problem is this solving?**
MotorolaNO environment is not rendering on the page. We receive an error message stating that the language (nn-NO) is not supported.


sheilagomes (approved):
LGTM

